PR #4, opened 2015-06-25
First installation fixes and .gitignore update

ajira86:


ajira86:
Github has automatically added commit 896ba7f, so I have to create a branch for future pull request :/
I let you evaluate if it can help

ajira86:
I saw the bzignore file that have the same structure as a gitignore fyi

Ppjet6:
You should remove the composer.lock from the .gitignore file. See [composer's recommendations](https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-file).

ajira86:
@Ppjet6 :+1:

edhelas:
Thanks for the PR :) 

In system.tpl you cannot put quotes around the two values. The first one is a hash that need to be displayed correctly in Javascript, and the second one is a boolean :p So maybe we can write some test in the controller to fix this ?

edhelas:
Linked with the issue #8

edhelas:
The changes are now in trunk, see this commit https://github.com/edhelas/movim/commit/97f96daa038dcc07e63d12710bab9982b4d844af

PR #5, opened 2015-06-26
Fix the 'Movim' icon in the menu on mobile

Ppjet6:
Changed the icon for md-cloud-queue instead of md-menu


edhelas:
It will not be clear for the user to have the Movim logo in the open menu on mobile. Maybe we can put a <- arrow instead to show that he can click and fold-back the menu ?

Ppjet6:
On 2015/06/26, Jaussoin Timothée wrote:

> It will not be clear for the user to have the Movim logo in the open menu on mobile. Maybe we can put a <- arrow instead to show that he can click and fold-back the menu ?

Why would the icon be different from the desktop version? Is it more
difficult to understand on mobile?

As you know this happens on my netbook as well and not just on mobile,
(screen size 1024x600), so maybe that's a deeper problem and we should
rethink the whole design.

Or maybe I should just give up and accept nobody nowadays has screen
size that small.. :(

PR #16, opened 2015-07-28
Commiting composer.lock

Ppjet6:
Commiting composer.lock so dependencies stay sync with upstream (this repo). See [Composer's recommendations](https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-file).


edhelas:
The composer.lock is now part of the project ;)

PR #55, opened 2015-10-13
Better readability for browsers with dark themes.

Dragnucs:
Text entries are hard to read on browsers or desktops that use dark themes. Forcing the color to black solves the issue.

### DIFF
--- a/themes/material/css/form.css
+++ b/themes/material/css/form.css
@@ -96,6 +96,7 @@ form > div > textarea {
     padding-top: 5rem;
     width: 100%;
     background-color: transparent;
+    color: black;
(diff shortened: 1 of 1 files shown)


edhelas:
The current color is color: rgba(0, 0, 0, 0.87); which is the color defined in the Material Design documents. What is a "dark theme" ?

Dragnucs:
[Arc Dark](https://github.com/horst3180/Arc-theme) is an example of a dark theme.

Maybe the used material templated is buggy.

This color is not explicitly set to inputs and textareas so while dark themes tend to set the color of these elements to a lighter one, which is completly legit, websites setting only a background color, like movim, generates issues.

He who changes the background, should also change the foreground.

movim-bug

PR #183, opened 2016-05-16
daemon: listen on loopback only

LudovicRousseau:
The Movim daemon and the apache server are on the same computer.
Apache will redirect the websocket to ws://localhost:8080/ so the daemon
only has to listen to the localhost (loopback) interface.

The Movim daemon is then no more available from the Internet.

### DIFF
--- a/daemon.php
+++ b/daemon.php
@@ -49,7 +49,7 @@ $core = new Core($loop, $argv[1], $argv[2]);
 $app  = new HttpServer(new WsServer($core));
 
 $socket = new Reactor($loop);
-$socket->listen($argv[2], '0.0.0.0');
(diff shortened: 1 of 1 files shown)


LudovicRousseau:
Another improvement would be to use IPv6 instead of IPv4 loopback address.
Maybe "::1" instead of "127.0.0.1" would work.

This is my first Movim patch. Maybe I completely failed to understand how Movim is working. Just tell me.

LudovicRousseau:
I checked. It does also work with "::1" instead of "127.0.0.1".
IPv6 is the future. Use it :-)

edhelas:
Seems ok for me to go for IPv6. I think that there is no issue on any OS to have this ?

PR #199, opened 2016-06-03
Increase requestURL timeouts

bremensaki:
Movim API calls using requestURL set the timeout to 1 second. This makes it nearly impossible to interact with the API from Australia due to latency.

Timeouts have been increased to 10 seconds in every case found.


edhelas:
Looks fine to me, can you please try with 3 or 5 sec. 10 sec seems a bit too much for me.

bremensaki:
I'll remove this whole PR, I didn't mean to include the typo fix in the same go, I meant to do that on a separate branch. Will talk to you in the chatroom when you get a moment.

PR #238, opened 2016-09-01
HTTP Upload fixes accroding to latest XEP-0363.

pztrn:
Latest XEP-0363 says that server should return "201 Created" response
code on successful file upload. See http://xmpp.org/extensions/xep-0363.html#upload.


edhelas:
Thanks !

Issue #123, opened 2016-01-27
Auto-focus on text input field in chat


Nyco:
GIVEN the browser window is in the background
AND the Discussions/Conversations/Chats tab is selected
WHEN the browser window gets the focus
THEN the keyboard pointer should already be positionned/focussed on the text input field

PR #239, opened 2016-09-02
querySelector workaround for chats with contacts which jid starts with numbers

pztrn:
querySelector requires some additional handling (e.g. escaping) when
ID starts from numbers. For example, if you have contact with jid
123@example.org - history WILL NOT load, because it will whine
about something like:

movim_tpl.js:86 Uncaught SyntaxError: Failed to execute 'querySelector'
on 'Document': '#123-example-org-chat-item' is not a valid selector.

With prepending "chat" to chat items we will workaround it without
need of additional escaping and checking like described here:
http://stackoverflow.com/questions/20306204/using-queryselector-with-ids-that-are-numbers

### DIFF
--- a/app/assets/js/movim_utils.js
+++ b/app/assets/js/movim_utils.js
@@ -42,7 +42,7 @@ var MovimUtils = {
         return r;
     },
     cleanupId: function(string) {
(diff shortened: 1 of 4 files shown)


edhelas:
Looks fine to me :)

PR #240, opened 2016-09-02
Pictures preview in MUC.

pztrn:
This PR enables pictures preview in MUCs.

How it looks like:

Picz preview in MUCs


edhelas:
Your patch was appending the pictures after each messages (li tags) which was syntactically wrong. I fixed it in this commit https://github.com/movim/movim/commit/dcce6252963c61d07c59166b05da657ddf0a2d7e .

PR #335, opened 2017-01-18
German transl fix

elouin:
Fixed one mistake and added some missing things to the german translation.


edhelas:
I've updated the German translation in Transifex, these will be merged in master soon.

PR #378, opened 2017-04-02
Nice UI for adding a contact through a Gateway

### LINKED ISSUE #367: Support XEP-0100 jabber:iq:gateway protocol
https://xmpp.org/extensions/xep-0100.html#addressing-iqgateway

This enables adding contacts through gateways with a very nice UI similar to:

I realize gateways are probably not high-priority for movim, but I am willing to do the coding for this.  Just want to make sure there's a chance the PR would be accepted before I put in the work :)

singpolyma:
Implements https://xmpp.org/extensions/xep-0100.html#addressing as well as a final fallback to the old %-encoding hack that some gateways still use for historical reasons.

Closes #367 

Depends on https://github.com/movim/moxl/pull/17

### DIFF
--- a/app/widgets/Roster/Roster.php
+++ b/app/widgets/Roster/Roster.php
@@ -5,6 +5,7 @@ use Moxl\Xec\Action\Roster\AddItem;
 use Moxl\Xec\Action\Roster\RemoveItem;
 use Moxl\Xec\Action\Presence\Subscribe;
 use Moxl\Xec\Action\Presence\Unsubscribe;
+use Moxl\Xec\Action\IqGateway;
 
 class Roster extends \Movim\Widget\Base
 {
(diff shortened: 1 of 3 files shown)


singpolyma:
@edhelas this one is ready for initial review now, and is fully functional in my tests :)

app/widgets/Roster/Roster.php
```diff
@@ -64,6 +69,34 @@ function onRoster()
...
+            'Roster.addGatewayPrompt',
+            $packet->from,
+            ''.$packet->content->prompt,
+            ''.$packet->content->desc
+        );
+    }
+
+    function onIqGatewaySet($packet)
+    {
+        $form = json_decode($packet->content->extra);
```
edhelas: No need of this extra step if it's an array, what do you think?

edhelas (commented):
Looks great :)

app/widgets/Roster/Roster.php
singpolyma: I was getting an error about not serializing SimpleXML without this.

singpolyma: after other changes, this no longer happens, it seems.  So I've removed the json encoding again

singpolyma:
@edhelas I believe I have made all your suggested changes :)

PR #395, opened 2017-06-02
Support ad-hoc commands on contacts who support them

singpolyma:
Depends on https://github.com/movim/moxl/pull/18

### DIFF
--- a/app/views/contact.tpl
+++ b/app/views/contact.tpl
@@ -22,6 +22,7 @@
         <?php } else { ?>
             <aside>
                 <?php $this->widget('ContactData'); ?>
+                <?php $this->widget('AdHoc'); ?>
             </aside>
             <div>
                 <?php $this->widget('ContactHeader'); ?>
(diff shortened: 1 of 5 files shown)


singpolyma:
Because there was some concern about what this looks like if it's "always included", here are some screenshots of what it looks like.

Most contacts will look exactly the same as they do now:

screenshot - 06172017 - 10 18 14 am

But contacts that represent services will also show the available commands:

screenshot - 06172017 - 09 44 22 am

I think this UX is nicer that requiring a second click to get at the commands, and it won't affect most users at all since the page is exactly the same as now when there are no commands.  If we *did* require clicking a sidebar item to get at the commands, where should they go?  In a modal, I suppose?

@edhelas let me know what you think.  I am willing to change it to a sidebar button, or whatever you prefer, but thought I should show the current design so we're sure it is needed before I build it.

singpolyma:
Based on conversation with @edhelas I have updated this PR to put the commands in the sidebar instead of the main area.  Screenshot of it in action:



And here you see that normal contacts look the same as before:



And that the existing ad-hoc commands for the server also look the same as before:

PR #439, opened 2017-08-29
quick fix to hide the <p>...default captcha message...</p>

ghost:
and display an image with the captcha image instead.

the submit button has been forced in order to allow keyboard press 'enter' when form fulfilled


edhelas:
I'm sorry but this is not a really beautiful way to solve that issue.
Injecting HTML with inline CSS and JS is not super clean to me.

ghost:
I agree, this is a quick and dirty fix. 
I did not investigate more into the php code. apologize myself.
any **clue** might be helpful, trying to help and make things smarter than this first PR.

by the way, I didn't manage to deal with the 30s refresh page, specially from this sign-up page.
and that's a bit odd for new users (who are not fast enough to register).

PR #469, opened 2017-09-21
Attempt at using salted passwords

MarcelWaldvogel:
- Create/update using `password_hash(…, PASSWORD_BCRYPT)`
- Compare using `password_verify() || sha1()`; `sha1()`
  kept so that passwords set before will still be accepted.
  The backward compatibility code can be retired
  once all admins have updated their passords.

Recommendation to admins: Change the password, even to the same value
to benefit from the more secure storage.

Disclaimers:
1. I did not test this
1. The initial password is still 'password'; this probably should be randomly chosen at first start and output. Where is this initial password set?
1. I noticed that session IDs are `sha1()` of the user password (among other data). I did not change that, as I do not know how frequently the session IDs are generated (BCRYPT is deliberately much slower than SHA-1)


Trim:
Hello,

In my point of view, these changes aren't able to manage an update of the `PASSWORD_DEFAULT` value.

To solve this issue, I'll do that:

* add a database update to add a `passwordAlgorithm` SQL column beside the password value
* set this column to the PHP corresponding value for `SHA1`
* modify the AdminloginController.php to check user entered password against the saved algorithm (then, you can remove the "or check sha1" part)
* Inside the "ok, you are admin" block, first check if the algorithm saved in colon is different from the one of `PASSWORD_DEFAULT`. If that's the case, update the password and password algorithms with newly computed hash/salt and algorithm.

Although, I don't know how Movim code work and how it manages database updates.

MarcelWaldvogel:
@Trim The idea of today's salted password mechanisms is to start with an algorithm identifier, to uniquely identify how it has been generated. Therefore, `password_verify()` will can determine the algorithm, and other algorithm-specific values (for BCRYPT, this is the number of iterations and the salt), as explained [in the PHP manual entry for `password_verify()`](http://php.net/manual/en/function.password-verify.php).

However, I find it a good idea to update the SHA1 hash to a BCRYPT hash on first login after the update. However, I do not know the internals well enough to know how to store the value back.

Ppjet6:
Istr `mud.php` writes the password in the database, you could probably have a look in there.

https://github.com/movim/movim/blob/b06deb85e95dda9d67d65c36bb8fc528879936f6/src/Movim/Console/ConfigCommand.php#L81-82

Trim:
Oh, well, that's better than ecpected :)

So maybe, we could add sha1 algo as prefix of current password instead doing sha1 check every time a wrong password is entered.

edhelas:
Thanks a lot for the PR, and sorry for the delay.

PR #488, opened 2017-10-04
Account for self-hosted HipChat instances.

Natureshadow:
Self-hosted HipChat instances are (currently) hard-coded to the *.btf.hipchat.com domain, so it is easy to catch those as well to support their protocol extension to MUC cards.


edhelas:
Close, fixed in https://github.com/movim/movim/commit/79b151b84e7db4fecfa857db4ebc7ddc3b90564c

PR #513, opened 2017-10-20
Adding tests with Travis CI and free/libre tools

ghost:


ghost:
There are new conflicts. There are problems with PHPMD:

- `phpmd` of Debian 9 does not support operator `??` of PHP 7.
  https://lornajane.net/posts/2015/new-in-php-7-null-coalesce-operator
  
  It can be "fixed" with: (see `src/Movim/Daemon/Core.php` on line 229)
  
  ```
  if(isset($arr['path']) && !empty($arr['path']) && $arr['path'] != null)
  {
      return $arr['path'];
  }
  return false;
- In `src/Movim/Widget/Base.php` on line 98, the syntax with 3 dots seems not to be supported (`return __(...$args);`).
- It is the same in `src/Movim/Widget/Wrapper.php` on line 147 (`return $widget->$method(...$params);`).

ghost:
The changes should be put in just one commit (instead of the original one + merges). If you agree with the diff of the sum of the commits, I can remove them by a single one.

tcyrus:
Isn't that what the [squash and merge](https://blog.github.com/2016-04-01-squash-your-commits/) option is for? The issue is you (or your editor) removed newlines from some of the `.php` files which have since changed. Undoing the changes to those php files should prevent the need for unnecessary merges. Something like this should do the trick.

```
git checkout 92f936e79f7be4b8552012a1054d5324dc413cd1 -- mud.php
git commit
```

edhelas:
Closing this PR. I might add proper code check later.

PR #522, opened 2017-10-28
Adding an EditorConfig file

ghost:


edhelas:
Is this usefull ? In which context ?

ghost:
It could be useful, especially with developers on Windows. Some years I did student project, most guys were on Windows, they produced files with CRLF (newlines) and with ISO 8859-15. Mixing encoding and/or how to do newlines in a project is weird and not coherent. Moreover some editors have no problem with mixing: I discovered files with CRLF newlines but also LF newlines, `git diff` can produce "noise" related to that.

PR #527, opened 2017-11-03
Minor changes in PHP source code

ghost:


app/controllers/AdminloginController.php
```diff
@@ -3,6 +3,16 @@
 
 class AdminloginController extends Base
 {
+    private static function getHashedPassword($password)
+    {
+        return sha1($password);
+    }
+
+    private static function isCorrectPassword($exceptedPassword, $givenPassword)
```
ghost, line 11: Would be useful to make https://github.com/movim/movim/pull/469 more readable.

Ppjet6:
Hi there!

Your changes look helpful indeed, thanks! I was just wondering if you could maybe try to split them into meaningful commits, instead of putting everything in the same bag.
Almost none of these changes look semantically related to me, maybe that could be a nice addition to your next PR :)

edhelas:
Indeed, there's some things in those PR that I agree and others that would need discussion. So I think it's better to split all those changes in smaller uniform chunks, if possible :)

ghost:
When I do minor changes, I prefer to do one commit instead of even smaller ones. But as you prefer the contrary, I will split.

PR #528, opened 2017-11-03
Minor modifications in app/models/

ghost:


app/models/message/Message.php
```diff
@@ -123,11 +128,7 @@ public function set($stanza, $parent = false)
                 $this->body = trim(html_entity_decode($this->body));
             }
 
-            if($stanza->markable) {
-                $this->markable = true;
-            } else {
-                $this->markable = false;
-            }
+            $this->markable = $stanza->markable;
```
edhelas, line 131: Here it check if the XML tag is there, $stanza->markable is not a boolean

app/models/contact/Contact.php
```diff
@@ -302,7 +302,7 @@ function getAlbum()
         $uri = str_replace(
```
edhelas: You can simply remove the getAlbum() methode, it's not called anywhere anymore

ghost: Ok, I will.

app/models/message/Message.php
ghost: What is the type of `$stanza`? I see no method `attributes` for [DOMElement](https://secure.php.net/manual/en/class.domelement.php).

edhelas:
Most of those changes are already in master now :)

ghost:
I rebased if you would like to integrate the rest.

PR #530, opened 2017-11-04
Removing some useless else and elseif (in PHP)

ghost:
A split and mix of https://github.com/movim/movim/pull/527 and https://github.com/movim/movim/pull/528

### DIFF
--- a/app/helpers/StringHelper.php
+++ b/app/helpers/StringHelper.php
@@ -84,24 +84,22 @@ function addUrls($string, $preview = false)
                             $content.
                             '</a>'
                         );
-                    } else {
-                        return stripslashes(
-                            '<a href=\"'.
-                            Route::urlize('contact', $link).
-                            '\">'.
-                            $content.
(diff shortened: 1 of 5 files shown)


app/helpers/StringHelper.php
```diff
@@ -84,24 +84,22 @@ function addUrls($string, $preview = false)
...
+                        Route::urlize('contact', $link).
+                        '\">'.
+                        $content.
+                        '</a>'
+                    );
+                }
+                if(filter_var($content, FILTER_VALIDATE_URL)) {
                     return stripslashes('<a href=\"'.$content.'\" target=\"_blank\" rel=\"noopener\">'.$content.'</a>').
                             ($lastTag !== false ? $lastTag : '');
-                } else {
```
edhelas, line 99: Those following changes has been tested?

ghost: No, why would it potentially break?

PR #531, opened 2017-11-04
Improving PHP class AdminloginController

ghost:
A split of https://github.com/movim/movim/pull/527, a bit improved

### DIFF
--- a/app/controllers/AdminloginController.php
+++ b/app/controllers/AdminloginController.php
@@ -3,6 +3,31 @@ use Movim\Controller\Base;
 
 class AdminloginController extends Base
 {
+    private static function getHashedPassword($password)
+    {
+        return sha1($password);
+    }
+
+    private static function isCorrectPassword($exceptedPassword, $givenPassword)
+    {
+        return $exceptedPassword == self::getHashedPassword($givenPassword);
+    }
+
+    private static function isCorrectCredentials($config, $credentialsGiven)
+    {
+        return (
+           isset($credentialsGiven['username'])
+        && isset($credentialsGiven['password'])
+        && $config->username == $credentialsGiven['username']
+        && self::isCorrectPassword($config->password, $credentialsGiven['password'])
+        );
+    }
+
+    private static function isCorrectCredentialsFromPost($config)
+    {
+        return self::isCorrectCredentials($config, $_POST);
+    }
+
     function load()
     {
         $this->session_only = false;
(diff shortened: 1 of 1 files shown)


app/controllers/AdminloginController.php
```diff
@@ -3,6 +3,31 @@
...
+           isset($credentialsGiven['username'])
+        && isset($credentialsGiven['password'])
+        && $config->username == $credentialsGiven['username']
+        && self::isCorrectPassword($config->password, $credentialsGiven['password'])
+        );
+    }
+
+    private static function isCorrectCredentialsFromPost($config)
+    {
+        return self::isCorrectCredentials($config, $_POST);
```
edhelas, line 28: There's a getPost method in the parent, maybe we should use that?

app/controllers/AdminloginController.php
```diff
@@ -3,6 +3,31 @@
 
 class AdminloginController extends Base
 {
+    private static function getHashedPassword($password)
+    {
+        return sha1($password);
+    }
+
+    private static function isCorrectPassword($exceptedPassword, $givenPassword)
```
edhelas, line 11: This looks a bit much no? Creating one line methods that are called only once?

app/controllers/AdminloginController.php
```diff
@@ -3,6 +3,31 @@
 
 class AdminloginController extends Base
 {
+    private static function getHashedPassword($password)
```
ghost, line 6: This may be more complicated in the future if a salt is added.

app/controllers/AdminloginController.php
ghost: SHA1 is not considered to be safe anymore.
https://en.wikipedia.org/wiki/SHA-1#Attacks
https://linuxfr.org/users/gouttegd/journaux/et-paf-le-sha-1

So it would be useful to change that in the future. If it is done, this method would be a bit longer: it would have to check the new hashing way and the old one (with SHA1) for compatibility.

app/controllers/AdminloginController.php
ghost: `cd app/ && grep -r 'class Base'` returns nothing. Is it in Modl, Moxl, or somewhere else? Edit: it does not seem to be in Modl or Moxl.

edhelas:
With the merge of this PR https://github.com/movim/movim/pull/469, I'm now closing this one

Issue #550, opened 2017-12-11
ability to post comments on non-movim.eu posts


hackervera:
right now it seems like any non-movim.eu account has commenting automatically disabled. if we create a a node on pubsub.non-movim.eu-domain.com with publish_permission = open we can allow people to publish comments for the post

mitchellurgero:
This is completely specific to the xmpp server, not movim (movim can manage the permissions but that is all done via xmpp, not movim directly)

Changing the permissions on the community while creating the node OR allowing it via say ejabberd.yml would resolve this for you friend.

hackervera:
What I'm suggesting *is* movim specific. I was suggesting that we alter the way we talk to pubsub node so that non-movim.eu domain posts can have comments enabled.

mitchellurgero:
Ohhh You mean on movim.eu specifically. 

Although this repo is for movim, it is not specific to movim.eu's setup, and what you are asking for is not specific for movim's code base either. I would recommend contact Edhelas deirectly about this.

hackervera:
Movim creates a node on comments.movim.eu, where does it get that value if not from the software?

edhelas:
https://github.com/movim/movim/blob/master/app/models/caps/CapsDAO.php#L71 if comments.server.tld exists

hackervera:
Why comments.server and not pubsub.server? movim is the only server i've seen with a comments pub so the fail rate seems like it would be pretty high.

mitchellurgero:
XMPP supports any subdomain for pubsub... has nothing to do with movim (movim just interfaces with the configured pubsub server of the users XMPP server.)

hackervera:
> XMPP supports any subdomain for pubsub... has nothing to do with movim

`comments.$server` is literally hard coded in

mitchellurgero:
> comments.$server is literally hard coded in

Sure it can be hard coded, I am just saying it's not movim specific, any XMPP client can use pubsub (what is being called ```comments.$server``` in the code. ).

As an example of what I mean - my xmpp server currently uses pubsub.domain.com, irc.domain.com, uploads.domain.com, and a couple other "pubsub" services are also thrown in there for organization, movim and any other xmpp client can use these services. Movim is just an XMPP client, it is not a server in the sense that it itself is serving xmpp requests.

That being said, I am not trying to argue against your original post, just stating that it's kind of in the wrong place is all.

edhelas:
[...]

PR #551, opened 2017-12-12
Update CapsDAO.php

hackervera:
Issue #550


edhelas:
Before jumping into conclusions, please check how the code is working fully please.
https://github.com/movim/movim/blob/master/app/widgets/Publish/Publish.php#L289

If Movim finds a Pubsub service called "comments." he creates the comments node there, if not, it fallback to the same service where the item is Published.

This behavior is there for a reason, it's to prevent to create thousands of comments node on the same services as the default nodes and use a lot of resources each time we try to discover those services.

PR #652, opened 2018-05-19
Font issues

mirabilos:
- split opensans.css from fonts.css so we have a single separate CSS file for each webfont used
- add licence information for both embedded fonts 
- use the original SVG from Material-Design-Iconic-Font instead of a whitespace-challenged one

While I am not a lawyer and this is no legal advice, I *believe* that, with this, Movim upstream is in compliance with the licences of the fonts as much as possible, given https://github.com/zavoloklom/material-design-iconic-font/issues/122 exists.

One thing we almost certainly (but not before upstream provides us enough information to actually do by fixing the aforementioned issue) **will** need to do is to put the actual terms of the specific MIT licence variant the Material-Design-Iconic-Font CSS uses, and their copyright notices (when they add them), in here as well. Until then, there’s nothing we can do since we can only retain what’s already there.


edhelas:
I'm planning to replace i with http://google.github.io/material-design-icons/#icon-font-for-the-web asap :) Thanks for the follow up.

edhelas:
The code has been updated ;)

mirabilos:
Thanks, but I see you reverted the split off `opensans.css` and a CSS for the icon theme.

It would help us in the distribution if you could bring that back, as we’ll have to replace them with the respective files from the Debian packages (for OpenSans, nothing changes, and apparently @Natureshadow will need to package the Google font as well; I thought it was packaged in Debian already, but it isn’t, but it’s buildable, so we’ll have to split it off).

PR #681, opened 2018-07-18
Improve configuration handling

JKingweb:
These changes prevent Movim from failing awkwardly when supplied an incomplete configuration file, by simply providing a default configuration that the user's configuration overrides. The defaults provide an invalid database type, so the configuration file must minimally provide that much.


edhelas:
I prefer to have the admin filling this file manually better than trying to autofill things for him.
This is something that he should do only once and never really touch again.

PR #684, opened 2018-08-12
Use new, safe Electron API

dos1:
Once movim/movim_electron#17 lands, Movim should use new API exposed by Electron app that doesn't rely on giving full access to the user's machine to some code loaded from the Internet ;)

It's not very urgent though, as the patch for Electron app features a shim that allows older pods to still fully function safely.


edhelas:
Thanks a lot !

PR #695, opened 2018-08-16
Fix notification suppression when the window is out of focus on Chromium

dos1:
According to specification, blur and focus events are called on window object.
Only Firefox (and maybe IE?) fires them also on document.

This fixes also the Electron client.


edhelas:
Thanks a lot for this fix :) Indeed it was a dumb mistake.

PR #696, opened 2018-08-18
Fix going through MAM pagination in chats

### LINKED ISSUE #687: On new registration, history from MAM is truncated
I've just did a fresh installation of newest Movim and started some chats. I noticed that, although the history did fill up, the newest message available was from May. Subsequent reconnections eventually showed some newer messages, but only single ones from later dates, clearly cut out of context.

I've that already when installing v0.13 earlier. When the pod was running, history was gathered properly - seems it's just the process of bootstraping it from MAM that's buggy.

Server: ejabberd 18.06

dos1:
Fixes #687

### DIFF
--- a/lib/moxl/src/Moxl/Xec/Action/MAM/Get.php
+++ b/lib/moxl/src/Moxl/Xec/Action/MAM/Get.php
@@ -51,11 +51,14 @@ class Get extends Action
         if(isset($stanza->fin)
         && isset($stanza->fin->set) && $stanza->fin->set->attributes()->xmlns == 'http://jabber.org/protocol/rsm'
         && isset($stanza->fin->set->last)
-        && !isset($this->_jid)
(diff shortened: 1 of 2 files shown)


lib/moxl/src/Moxl/Xec/Action/MAM/Get.php
```diff
@@ -40,6 +40,18 @@ public function setBefore($before = true)
         return $this;
     }
 
+    public function setStart($start)
```
edhelas: I don't really get the need of those two methods, this piece of code is already taking care of that https://github.com/movim/movim/blob/master/lib/moxl/src/Moxl/Xec/Action.php#L50 :p

dos1: Ooooh. That explains a lot xD

PR #697, opened 2018-08-18
Fix all messages being marked as edited and quoted on SQLite

dos1:
For some reason, when using SQLite, those values reach JS land as string.
In JavaScript, "0" is a truthy value, so we need a conversion.

### DIFF
--- a/app/Message.php
+++ b/app/Message.php
@@ -16,6 +16,12 @@ class Message extends Model
         'type'    => 'chat'
     ];
 
+    protected $casts = [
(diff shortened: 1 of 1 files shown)


edhelas:
I'm not a big fan of that patch. I think you can tell Eloquent to cast the type when requesting stuff from the DB (https://mattstauffer.com/blog/laravel-5.0-eloquent-attribute-casting/) :)

edhelas:
Thanks!

PR #700, opened 2018-08-21
Fix disappearing contact entries after being renamed by another client.

dos1:
This actually modified a cached roster entry of another random user that
happened to have the same JID in their roster.

PS. I wonder if there may be other places where similar bug happens... :P


edhelas:
thanks for that one

PR #702, opened 2018-08-21
Respect <fin complete="true"> attribute on MAM results.

dos1:
Without that, Movim requests an additional, pointless query that will return nothing.

Turned out some servers even didn't handle that properly, resulting in an infinite loop of MAM requests.

BTW. Is this a better way to deal with boolean XML attributes?


dos1:
Also, that set/last check could be removed now, but not sure if `complete="true"` has been a part of every supported MAM specification and if every implementation returns it, so I left it untouched. [edit] it is, so I removed it now.

edhelas:
Please check the presence of the complete attribute as well, as this could fire a warning.

> If it is not the last page of the result set, the server MUST either omit the 'complete' attribute, or give it a value of 'false'.

PR #704, opened 2018-08-22
Handle MUC 110 status correctly if the stanza contains more than one <x> elements.

dos1:


dos1:
I'm afraid there may be more bugs like that. I think I just found one in SIMS handling.

edhelas:
Good finding indeed! But instead of looping over the `x` elements, I think you can use a xpath request there. The stanza is a `SimpleXML` object already.

PR #705, opened 2018-08-22
Fix urn:xmpp:mam:1 support

dos1:


edhelas:
Indeed the version needs to be passed along the requests as well.

Thanks.

PR #706, opened 2018-08-23
Verify shared picture size when proxying

dos1:


edhelas:
Do you think it's better this way 🤔 ?

edhelas:
Also be careful because this route is used at several places (see `protectPicture()`)

dos1:
I was making https://github.com/dos1/movim/commit/956383e31182e28cf819b9b7504275315f84ae45 and to test it, I have used first image files that appeared in the file dialog. Those turned out to be some 32x32 icons that didn't work because they were too small :P So I though it may make sense to replace that minimal size check with something more robust. I'll check out other places - I did some git grepping for them, but I could have missed it if it's constructed differently.

BTW. Is https://github.com/dos1/movim/commit/956383e31182e28cf819b9b7504275315f84ae45 something you would even consider merging? If not, I won't bother with PRing it :P

edhelas:
Hi !
After those 4 years of silence I'm checking up this PR. Would it be possible to talk about it by chat, I'm still figuring out the final goal of this PR.

Neustradamus:
@dos1: Have you seen the last @edhelas comment?

edhelas:
The picture proxy was recently refactored, this should not be relevant anymore

PR #716, opened 2018-09-18
Make bubble icons non-selectable

dos1:


edhelas:
But why?

dos1:
It just looks like something broken when you select text from multiple balloons :P

edhelas:
And you actually need to click on those icons in some places, like in the MUC widget.

dos1:
...and `user-select` doesn't influence that in any way. It just makes the UI feel a bit more solid, as the placeholder text in the avatarless bubbles won't appear selected anymore.

edhelas:
Well thanks…

PR #717, opened 2018-09-18
Fix MUC joining when returned presence doesn't contain the same ID

### LINKED ISSUE #693: MAM on MUC doesn't work if there's other client with the same nick in room when Movim joins
I found an entry in the changelog for version 0.12:
```
* Add support of MAM (up to mam:2) for the MUCs
```

Well, it doesn't work :D Clean installation cleared the history of all the rooms, despite of it being available in MAM for at least some of them.

dos1:
It could happen for plenty of reasons, the main one being an other
resource already connected to that MUC under the same nick.

Fixes #693

### DIFF
--- a/lib/moxl/src/Moxl/Xec/Action/Presence/Muc.php
+++ b/lib/moxl/src/Moxl/Xec/Action/Presence/Muc.php
@@ -29,7 +29,7 @@ class Muc extends Action
         }
 
         // Save the state in the session to handle the callback later
-        $session->set($this->_to . '/' .$this->_nickname, true);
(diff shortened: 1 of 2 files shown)


Natureshadow:
Pretty please merge this for 0.14.2 ☺!

poVoq:
I guess this might be circumvented with the new multiple identities feature?

edhelas:
Properly fixed in the XMPP stanza Handler there https://github.com/movim/movim/commit/074c079443404dee023fccb62924e8d3ce13ba72

Took some times, but it's fixed :3

PR #721, opened 2018-09-24
Fix text representation of replaced emoji.

Natureshadow:
This makes it so that…
…with images turned off or unavailable, the character is displayed unchanged,
…when graphics are available, the text on mouse hover explains the emoji,
…when copy and pasting messages, the emoji character is copied verbatim.


edhelas:
Small change, good change. Thanks!

PR #722, opened 2018-10-01
Add name an description to composer.json.

Natureshadow:
From Debian work, where the dh_phpcomposertoolchain requires these fields.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,4 +1,6 @@
 {
+    "name": "movim/movim",
+    "description": "Decentralized social platform based on XMPP",
     "autoload": {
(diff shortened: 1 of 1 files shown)


edhelas:
Is "Responsive web-based cross-platform XMPP client" not better? I am wondering what could be the best description for Movim.

Natureshadow:
>Is "Responsive web-based cross-platform XMPP client" not better? I am
>wondering what could be the best description for Movim.

Don't know - I simply copied the description from your bug fat README headline :P.

edhelas:
Thats fine for me !

PR #732, opened 2018-11-14
Check for the correct XEP-0166 feature

linkmauve:
This allows interoperability with clients which respect the
specification and only expose the 'urn:xmpp:jingle:1' feature.

In the future, advertisement of the bogus feature should be removed,
but it is kept for compatibility with existing Movim for now.


edhelas:
Thanks a lot for finding this issue!

PR #748, opened 2018-12-16
Make public URL and WS port configurable from distribution environment file

Natureshadow:
This is needed so to change the public URL (and protocol used to access movim) the systemd unit does not have to be changed. In Debian, we will make this configurable through debconf questions.


edhelas:
Thanks for the PR, looks fine to me :)

PR #749, opened 2018-12-16
Log errors if interprocess communication fails

mirabilos:
This aids searching for errors. We’re currently tracking a 500 Internal Server Error being retuned by the API…

### DIFF
--- a/app/helpers/UtilsHelper.php
+++ b/app/helpers/UtilsHelper.php
@@ -553,6 +553,9 @@ function requestURL(string $url, int $timeout = 10, $post = false, bool $json = 
     }
 
     $content = curl_exec($ch);
+    if (curl_errno($ch) == 0) {
+        error_log("URL request for " . $url . " returned error: " . $content);
+    }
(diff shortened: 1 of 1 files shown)


mirabilos:
Eh, no, this is totally wrong. `curl_errno()` is 0 when _no_ error occurred.

Whyever an HTTP 500 is not reported as error is… weird.

edhelas:
I'd also prefer to rain an Exception as Movim can properly catch them https://github.com/movim/movim/blob/master/src/Movim/Bootstrap.php#L313 :)

mirabilos:
Hmh. The exception *is* raised, around line 300 of `StreamingServer.php`. I’m currently hoping it would contain more information about why chatting does not work.

mirabilos:
cURL is worse:

>  Note that response status codes which indicate errors (such as 404 Not found) are not regarded as failure. curl_getinfo() can be used to check for these. 

So, apparently, one has to check what `curl_getinfo($ch, CURLINFO_HTTP_CODE)` returns. Meh. Sorry about the noise.

PR #758, opened 2018-12-22
Fix migration for SQLite.

Natureshadow:
SQLite does not support dropping foreign keys. However, it will cleanly replace the foreign key later in the migration if ignored.

### DIFF
--- a/database/migrations/20180515205626_change_messages_primary_key.php
+++ b/database/migrations/20180515205626_change_messages_primary_key.php
@@ -8,7 +8,10 @@ class ChangeMessagesPrimaryKey extends Migration
     public function up()
     {
         $this->schema->table('messages', function(Blueprint $table) {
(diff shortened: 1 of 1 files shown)


edhelas:
Hi, thanks for the PR, is it possible to use the {} for the if condition? And leave a carriage return after the closing one :)

Natureshadow:
Sure, done. I don't think it makes sense here, but you make the styles ;).

Natureshadow:
OK, asking to use another style than used by yourself in the same file seems ridiculous ☺.

Natureshadow:
I propose you merge, then fix the style in the whole file (and probably a ton of other files). Having two styles mixed in the same file is worse than not following a style guide.

Natureshadow:
Oh, then again, the styles were already mixed before…

edhelas:
Danke :3

PR #759, opened 2018-12-22
Handle early exceptions

mirabilos:
If an exception is thrown before $this->loadHelpers() was called, the exception handler fails because truncate() is not yet defined. Problem reported by @Natureshadow.


Natureshadow:
Signing off the patch (tested against 0.14.1-rc2).

edhelas:
Thanks :)

PR #767, opened 2018-12-29
Update XMPPtoForm.php

sandeepbisht:
saveHTML returns HTML.
saveXML added XML definition line in return.


edhelas:
This was already fixed in master, but thanks for noticing it :)

PR #782, opened 2019-01-14
Update Help page

realkinetix:
I re-arranged the Help page contents and added in some FAQs to answer some questions that I had.  However, some information is hard coded, such as "chat.isurf.ca", and none of it is in any language files.  However, it is step one for me, so I'm creating the pull request and am looking forward to the suggestions from Timothée.


edhelas:
Thanks for the PR! If you didn't do any translations in the PR, I'd prefer to not include the .po as well. I prefer to have specific commits with those files in, in any cases Movim is falling back to the English ones defined in the .ini files :)

realkinetix:
I guess I should comment back!  heh...  Thanks for the information on how it should be done.  I hope the PR is now in appropriate shape?

edhelas:
I fixed and merged your branch :) A few comments:
- Be careful by not mixing tabs and spaces in the same file, Movim is only based on spaces :p 
- Please use intelligible keys for the translations
- You can put small icons in the lists to make it clear

PR #796, opened 2019-02-14
[WIP] Make Movim a Progressive Web App

### LINKED ISSUE #639: Add Progressive Web App metadata
Hi all,

I would like to use movim on the iPhone. Since there is no app I would be fine just using the site, added as a separate icon on my home screen (ad per „add to home screen“ in mobile Safari). There are a few small issues making the experience less than ideal:
- the home screen entry has no nice icon (there is some special Tag the site can use to provide one, I can look it up if you want)
- scrolling is weird
- safari always shows navigation controls (ideally iOS would open it „like an app, I.e separate from safari and without navigation controls, not sure what the site has to do to achieve this, but I can do some research...)

AMDG2:
From the user point of a view a progressive web app is a web page that you can "install" on your mobile (and later on your PC)

Making Movim a PWA will allow anyone to install it easily on its smartphone. Full PWA has push notifications and offline capabilities

This PR fixes  #639

Todo:

- [x] Make movim installable on smartphones
- [ ] Add service worker to allow installation from Chrome on Smartphone
- [ ] Proxify images and other contents to avoid non-https content served from Movim (this breaks the PWA experience by opening the app in browser)
- [ ] Add push notifications
- [ ] Add some offline capabilities

### DIFF
--- a/app/views/page.tpl
+++ b/app/views/page.tpl
@@ -12,6 +12,10 @@
         <link rel="icon" type="image/png" href="<?php $this->linkFile('img/app/48.png');?>" sizes="48x48">
         <link rel="icon" type="image/png" href="<?php $this->linkFile('img/app/96.png');?>" sizes="96x96">
         <link rel="icon" type="image/png" href="<?php $this->linkFile('img/app/128.png');?>" sizes="128x128">
+        <link rel="apple-touch-icon" href="<?php $this->linkFile('img/app/48.png'); ?>">
+        <link rel="apple-touch-icon" href="<?php $this->linkFile('img/app/96.png'); ?>" sizes="96x96">
+        <link rel="apple-touch-icon" href="<?php $this->linkFile('img/app/128.png'); ?>" sizes="128x128">
+        <link rel="apple-touch-icon" href="<?php $this->linkFile('img/app/512.png'); ?>" sizes="512x512">
         <script src="<?php echo BASE_URI; ?>scripts/favico.js"></script>
         <script src="<?php echo
             \Movim\Route::urlize('system') .


edhelas:
The webmanifest has been merged recently on master :)
You can see an example there https://nl.movim.eu/?manifest

AMDG2:
Ok, I will rebase then. Are you interested in the rest of the tasks ?

edhelas:
Depends which ones.
- Push notification is actually not for Movim I think but more on the XMPP server.
- Offline capabilities is not really possible regarding the Movim architecture.
- Proxyfying images is also limited by the architecture (but done on a certain level already).
- And I don't know what is the service worked thing.

AMDG2:
The idea is to implement a service worker. The service worker can brings the following:

- install as an app with Chrome
- push notification
- offline support

The push notifications is tricky, it allows the user to receive push notifications when Movim is not open in the browser. Movim should act as a gap between XMPP server and push server from different browser vendors (at the end you support push notification only for Chrome and Firefox for example) and the instance owner would have to register manually and optionaly.

Proxyfying is mandatory to avoid any http content served through the PWA. In firefox if any content is loaded through http, the application is closed and a new tab is opened at the same place.

Offline support can be:

- keep current post and conversations displayed and navigable
- allows editing drafts of posts locally
- load movim static from cache when browser is offline (strategy online first)

edhelas:
Hi, are you planning to do the other points on this PR :) 
We can just review and merge what's there and we'll see for the rest another time.

AMDG2:
Hi, I will check this this week or next week.

edhelas:
Ping :)

edhelas:
I made some work recently to make Movim a PWA. It is now detected as it on Android and Chrome desktop. I'm closing this PR ;)

PR #809, opened 2019-02-26
Restart movim daemon after failure.

Natureshadow:
The movim daemon reacts indignant to various situations, like the database backend being restarted in the wrong moment, and then exits. systemd can try to restart it automatically in that case.


Natureshadow:
Pretty please merge this for 0.14.2 ☺!

edhelas:
Looks fine to me, thanks for the PR :)

PR #826, opened 2019-04-28
Install pako library externally. Fixes #824.

### LINKED ISSUE #824: Foreign code without source form and copyright
Since 0.14.2rc4, the source tree contains a compiled form of the Pako javascript library, but it does not contain the source form of that code and does not mention the copyright and licence for the copied code. The source of Movim thus is currently illegal and not distributable.

Natureshadow:
Pako is difficult to distribute in source form, both for its modular
structure and its different licences. It is hard enough to distribute
legally as a separate package, so let that be someone else's problem ;).
We integrate npm (for installing JS assets) with composer through the
foxy composer plugin.


edhelas:
Wow damn, please, let's try to avoid npm at all means.
If there's no alternatives I'd prefer to avoid having pako at all than this package dependency system.

Natureshadow:
I totally agree with you there :rofl:. However, having just the compiled pako.js in the repository is definitely not an option either. You would have to copy the whole source tree into the movim sources, including all the different copyright and licence notes (it is, in contrast to what the pako maintainers state in their metadata, a mix of different Zlib, BSD and MIT licences).

Of course, not having another JS library at all would be the best solution, but you probably chose pako for a reason. Yet, maybe driving forward the permessage-deflate support in Ratchet so the WebSocket can use the compression support included in browsers would also be an option.

edhelas:
Yeah I'll consider the second one as well. See if I can just get rid of that library and simply use the browsers features, it's not that Movim will stop working if this compression is not enabled everywhere.

https://github.com/movim/movim/blob/master/app/assets/js/pako.js is not the compiled version no ?

Natureshadow:
> https://github.com/movim/movim/blob/master/app/assets/js/pako.js is not the compiled version no ?

Yes, it is.

Natureshadow:
Also, permessage-deflate works in all browsers Movim is supported in. It is the server support in Ratchet which is missing.

Natureshadow:
Seems permessage-deflate in Ratchet is on its way and will be available in the near future: https://github.com/ratchetphp/RFC6455/pull/23

So I suggest dropping pako.

drbh:
https://github.com/drbh/wasm-flate is faster then pako and is a standalone WASM binary. You can test the speed difference [here](http://wasm-flate.s3-website-us-east-1.amazonaws.com/)

It's a bit bigger than pako at ~90kbs uncompressed, compared to pakos ~45kbs minified. Dont know is size is an issue in your use case. 

WASM can be run inline in PHP by instantiating the binary using a WebAssembly runtime like wasmer. PHP instructions are [here](https://wasmerio.github.io/php-ext-wasm/wasm/)

Hope this is helpful!

Issue #829, opened 2019-04-29
Please split cache/ directory


Natureshadow:
The `cache/` directory is used for internal things like sockets, template caching, etc., thus, it should not be readable by the webserver (or not delivered). The example nginx config disables access to `/cache/`. But this breaks Movim because the directory is also used for caching of avatars (and probably more), which are directly delivered from there.

Please split the directory into one for cached content to be delivered to the browser and one for internal caching.

edhelas:
Done there https://github.com/movim/movim/commit/323a3bc78cd5ef5da9a5b2e161a37dd5f6225536

PR #830, opened 2019-04-29
Allow access to *.jpg in cache/ directory for nginx.

Natureshadow:
This works around #829.


edhelas:
You can authorize the whole directory pretty much

edhelas:
You don't need that anymore since this commit that moves all the public cache under the public/ directory https://github.com/movim/movim/commit/323a3bc78cd5ef5da9a5b2e161a37dd5f6225536

PR #831, opened 2019-04-29
Implement ACL for /cache and /log in apache config.

Natureshadow:


edhelas:
I think the whole cache folder can be authorized

edhelas:
Regarding the recent changes in Movim, this is not relevant anymore

Issue #835, opened 2019-05-04
Movim replaces garlic emoji with smiley picture


mirabilos:
When I type 🧄 (U-0001F9C4) in a chat, Movim replaces it with random large (about 128x128px) smlley images — interestingly enough, not the same… the first was https://chat.teckids.org/theme/img/emojis/svg/1f600.svg and the second was https://chat.teckids.org/theme/img/emojis/svg/263a.svg õÕ

It seems to me I’ll have to look at the replacement code again…

mirabilos:
Just received confirmation from @Natureshadow that it also makes a foot into a cake.

I’m not sure how much of that code is still mine, but I’ll have a look as time permits, so feel free to assign this issue to me.

debdolph:
If i send a foot emoji via Movim, a cake appears in the chat box. But in the preview bar at the left the foot is rendered correctly.
![Screenshot_20190505_161944](https://user-images.githubusercontent.com/25152803/57195372-e6fec080-6f51-11e9-868b-02e4085963de.png)
![Screenshot_20190505_162028](https://user-images.githubusercontent.com/25152803/57195373-e9f9b100-6f51-11e9-951f-1dff995f50fa.png)

mirabilos:
Tom Teichler dixit:

>If i send a foot emoji via Movim, in the chat box a is cake displayed.
>But in the preview bar at the left the foot is rendered correctly.

OK, thanks for the extra info. I’ll have a look as time permits.

debdolph:
Once you're there ;)... can you please add a dolphin emoji :D? 🐬

mirabilos:
Tom Teichler dixit:

>Once you're there ;)... can you please add a dolphin emoji :D? 🐬

Emoji are added to Unicode. What Movim does here is to replace the
emoji with pictures. Those are up to edhelas, I’ve just fixed the
code (and will fix it again) that does this replacing.

Besides, garlic is more important… :þ

Natureshadow:
>>Once you're there ;)... can you please add a dolphin emoji :D? 🐬>
>>
>Emoji are added to Unicode. What Movim does here is to replace the>
>emoji with pictures.

There *is* a dolphin emoji (which you can see by looking at the dolphin emoji in your mail).

The request probably is about the emoji drawer.

mirabilos:
[...]

PR #837, opened 2019-05-08
Possibly fix the Emoji substitution errors

mirabilos:
This is the likely fix for https://github.com/movim/movim/issues/835 plus a couple of code cleanup activities in this forgotten area of the code, and it extends the testsuite for the replacement code to validate that all *other* codepoints pass through unchanged.

Pending testing…

### DIFF
--- a/src/Movim/Emoji.php
+++ b/src/Movim/Emoji.php
@@ -1,7 +1,7 @@
 <?php
 /*-
- * Copyright © 2018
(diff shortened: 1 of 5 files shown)


mirabilos:
@DummerDelfin says this fixes the issue.

edhelas:
Looks good to me, and indeed it fixes the related issue

PR #843, opened 2019-05-23
[WIP] Fix and complete server-side JID validation and escaping.

Natureshadow:


edhelas:
Thanks a lo, I'll review it asap!
You've added a dependency, is it OK regarding the packages (like the Debian one?).

Natureshadow:
Hi,

> Thanks a lo, I'll review it asap!
> You've added a dependency, is it OK regarding the packages (like the Debian one?).

For Debian, I will of course take care of that.  For everything else, it has
to be - no JID validation is possible without that (except with copying the
code, of course).

app/helpers/StringHelper.php
```diff
@@ -131,7 +136,14 @@ function getHashtags($string): array
  */
 function echapJid($jid): string
 {
-    return str_replace(' ', '\40', $jid);
+    $jid_parts = explodeJid($jid);
+
+    $from = array('\\',   ' ',    '"',    '&',    '\'',   '/',    ':',    '<',    '>',    '@');
```
edhelas, line 141: Looks great but I'll have to check if it doesn't break stuff in the UI

app/helpers/StringHelper.php
```diff
@@ -131,7 +136,14 @@ function getHashtags($string): array
...
+    $to =   array('\\5c', '\\20', '\\22', '\\26', '\\27', '\\2f', '\\3a', '\\3c', '\\3e', '\\40');
+
+    $jid_parts['username'] = str_replace($from, $to, $jid_parts["username"]);
+
+    return ($jid_parts['username'] !== '' ? $jid_parts['username'].'@' : '').$jid_parts['server'].($jid_parts['resource'] !== '' ? '/'.$jid_parts['resource'] : '');
```
edhelas: Could you please re-indent that part ? Like
```
$whatif
    ? 'yeah'
    : 'nope';
```

app/helpers/StringHelper.php
```diff
@@ -392,3 +412,86 @@ function urilize($path, bool $noTime = false): string
 function implodeCsv($value) {
     return implode(', ', $value);
 }
+
+/**
+ * Returns true if the value is an IPv6 or IPv4 address (in that order)
+ */
+function isIpAddress($value) {
```
edhelas: Can you please put a carriage return to all the `function() {` in that file ? :3

app/helpers/StringHelper.php
```diff
@@ -392,3 +412,86 @@ function urilize($path, bool $noTime = false): string
...
+ * or false if it is invalid under that profile.
+ */
+function prepJidPart($value, $prep) {
+    try {
+        $ret = $prep->apply($value);
+    } catch (ProfileException $e) {
+        return false;
+    }
+
+    if ((strlen($ret) < 1) || (strlen($ret) > 1023)) {
```
edhelas, line 453: 1023, magic number ?

app/helpers/StringHelper.php
```diff
@@ -392,3 +412,86 @@ function urilize($path, bool $noTime = false): string
...
+ * Returns a JID, stringprep'ed with the given profile,
+ * or false if it is invalid under that profile.
+ */
+function prepJid($value) {
+    // Split JID into parts
+    $jid_parts = explodeJid($value);
+
+    // Validate local part against Nodeprep profile
+    if ($jid_parts['username'] !== '') {
+        $jid_parts['username'] = prepJidPart($jid_parts['username'], new Nodeprep());
```
edhelas: I prefer to not put parenthesis if you don't have parameters (`new Nodeprep`). But that's ok like that :p

app/helpers/StringHelper.php
```diff
@@ -392,3 +412,86 @@ function urilize($path, bool $noTime = false): string
...
+    // Validate resource part against Resourceprep profile
+    if ($jid_parts['resource'] !== '') {
+        $jid_parts['resource'] = prepJidPart($jid_parts['resource'], new Resourceprep());
+        if ($jid_parts['resource'] === false) {
+            return false;
+        }
+    }
+
+    // Finally, validate domain part (IPv6, IPv4, FQDN, in that order)
+    if (!isIpAddress($jid_parts['server'])) {
```
edhelas: Please be consistent, could you change all the variables to camelCase instead of snake_case. 
I prefer to use camelCase in Movim :)

app/widgets/Chat/Chat.php
```diff
@@ -1014,7 +1012,7 @@ private function prepareComposeList(array $list)
      */
     private function validateJid($jid)
     {
-        return (Validator::stringType()->noWhitespace()->length(6, 60)->validate($jid));
+        return prepJid($jid);
```
edhelas: You can simply remove that method and directly call prepJid ;) Same with the others.

edhelas (commented):
Thanks a lot <3
I've put a few comments, mostly formating and cleaning. I'll do a live test of the branch after that to see if everything looks fine.

app/helpers/StringHelper.php
Natureshadow: When displaying a JID, you SHOULD translate all the characters back, says the standard.

app/helpers/StringHelper.php
Natureshadow: What do you mean?

app/helpers/StringHelper.php
Natureshadow: I learned on a Java type-writer :P

app/helpers/StringHelper.php
Natureshadow: Did that for everything I touched. Fixing that style for your own code should go outside this PR ;).

Natureshadow:
I updated a lot, please review again, @edhelas ☺!

edhelas:
As I'm not planning to maintain a specifical stringprep dependency from that I'm assuming that the only proper way to do so would be to directly call "idn" (http://www.gnu.org/software/libidn/manual/html_node/Invoking-idn.html) and call it from PHP. I'm already using it during the registration process (see https://api.movim.eu/accounts/register) and it has a complete Nodeprep implementation.

edhelas:
Let's see if I can revive this PR :)

edhelas:
Movim is now having a JID validation system, I'm planning to generalize it in an upcoming refactor

PR #898, opened 2019-12-31
Upgrade illuminate/database to 6.x

rgson:
Hi,

I've attempted to upgrade the `illuminate/database` dependency to version 6.x.

I heard from the Debian package maintainers that Movim is incompatible with newer upstream versions of illuminate/database. I'd like Debian to pack a 6.x version of the Illuminate packages, so this is my attempt to change that.

Judging from the [official migration guide](https://laravel.com/docs/6.x/upgrade) (and a light skimming of the [diff](https://github.com/illuminate/database/compare/5.8...6.x?diff=split&w=1)), I believe the required changes to be fairly minor: the main one being the declaration of primary key types.

> [Version] 6.0 has received performance optimizations for integer key types.
> If you are using a string as your model's primary key, you should declare
> the key type using the `$keyType` property on your model.

I've thus gone ahead and added the `$keyType` property to models using non-integer keys, based on what I could tell from the models themselves and the migrations. For the custom model class with composite key support, I made up the `"composite"` key type, just to replace the default.

Other differences from 5.8 to 6.x that seem inconsequential:
- Requires PHP 7.2
  => Movim already requires 7.3
- Requires Carbon 2.x instead of 1.x
  => 2.x was already in `composer.lock`
- The `Illuminate\Database\Capsule\Manager` class' `table` method changed signature
  => Never used with more than the first argument anyway
- The cursor method now returns an instance of `Illuminate\Support\LazyCollection`
  => Never used
- The `BelongsTo::update` method changed behavior
  => Never used
- `toArray` will now cast `Arrayable` attributes to an array
  => Presumably harmless

I believe this to be ready, but **I've been unable to test it**. I don't have an environment with a working Movim setup where I can test it.

This thus needs some assistance before it can be considered for merging, preferably from someone who knows enough about Movim to test it properly.

Thanks.

### DIFF
--- a/app/Contact.php
+++ b/app/Contact.php
@@ -11,6 +11,7 @@ use Movim\Picture;
 class Contact extends Model
 {
     protected $fillable = ['id', 'nickname', 'mood'];
+    protected $keyType = 'string';
     public $incrementing = false;
 
     public function user()
--- a/app/EncryptedPassword.php
+++ b/app/EncryptedPassword.php
@@ -7,6 +7,7 @@ use Illuminate\Database\Eloquent\Model;
 class EncryptedPassword extends Model
 {
     protected $fillable = ['id'];
+    protected $keyType = 'string';
     public $incrementing = false;
 
     public function session()
--- a/app/Invite.php
+++ b/app/Invite.php
@@ -7,6 +7,7 @@ use Illuminate\Database\Eloquent\Model;
 class Invite extends Model
 {
     public $primaryKey = 'code';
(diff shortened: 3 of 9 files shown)


edhelas:
Hi, thanks for the feedback, I'll try to have a look at it asap :)

edhelas:
Hey, I've tested your changes, everything seems to work perfectly.
It's now merged :) Thanks a lot !

rgson:
Great! Thanks for testing and reviewing it.

PR #970, opened 2021-02-08
Fix select menus white-on-white in Night mode

ltguillaume:
Fixes white text on white background issue in drop-down menus of <select> elements.


edhelas:
Thanks :)

PR #973, opened 2021-02-11
More consistent (Shift+)Enter behavior (newline or send)

ltguillaume:
Let the behavior depend on whether the user is actively using a touch screen, instead of letting the screen width be the determining factor.

This should also take into account that some people have touchscreen devices, but use keyboard+mouse. In such a case, default desktop behavior (Enter = send) will be used.


edhelas:
Thanks :)

PR #982, opened 2021-03-20
Handle XEP-0353, Jingle Message Initiation, In Carbons

### LINKED ISSUE #980: Handle Jingle Proposals in Carbons
While testing some things I had Gajim open. My Gajim has a priority greater than 0, and so all messages sent to me during that time were sent to Gajim, but were forwarded to my Movim. Everything is working fine.

But, when a jingle call came in, the [Jingle Message Initiation](https://xmpp.org/extensions/xep-0353.html) message was sent to Gajim, which didn't know anything about it, and only forwarded to Movim. This didn't cause any ringing, and the caller just thought I hadn't picked up.

To clarify, here's what Movim received:

```xml
<message to='[redacted]' from='[redacted]' type='chat'>
  <received xmlns='urn:xmpp:carbons:2'>
    <forwarded xmlns='urn:xmpp:forward:0'>
      <message to='[redacted]' xmlns='jabber:client' from='[redacted]' id='proposal%215be5e3740f1d48' type='chat'>
        <propose id='215be5e3740f1d48' xmlns='urn:xmpp:jingle-message:0'>
          <description media='audio' xmlns='urn:xmpp:jingle:apps:rtp:1'/>
        </propose>
        <stanza-id xmlns='urn:xmpp:sid:0' by='[redacted]' id='FUlsaSEOfjK2iI1a'/>
      </message>
    </forwarded>
  </received>
</message>
```

psycotica0:
If all clients have the same priority, then these jingle messages work
fine. They're sent to all clients, and everyone rings, whoever picks up
tells the others to stop ringing, etc.

But if one client has higher priority, then only that client will get
these jingle messages directly, and the other clients will just get a
carbon informing them of the messages.

That means that, in practice, if I have Movim and Gajim running and
someone sends a Jingle Message because Movim supports those, then Movim
won't ring. But Gajim won't know what the jingle messages are, and so I
won't be told in any way that someone's trying to call me, and they'll
think I've seen their call and I'm just ignoring them.

This implementation looks in the carbons for messages related to this
XEP, and if there are any then we handle them as though we had received
them directly.

(Closes #980)

### DIFF
--- a/lib/moxl/src/Xec/Payload/Carbons.php
+++ b/lib/moxl/src/Xec/Payload/Carbons.php
@@ -27,6 +27,8 @@ class Carbons extends Payload
                 // Another client just displayed the message
                 $displayed = new Displayed;
                 $displayed->handle($message->displayed, $message);
+            } elseif (count($jingle_messages = $stanza->xpath('//*[@xmlns="urn:xmpp:jingle-message:0"]')) >= 1) {
+                \Moxl\Xec\Handler::handleNode($jingle_messages[0], $message);
             }
         }
     }
--- a/lib/moxl/src/Xec/Payload/Carbons.php
+++ b/lib/moxl/src/Xec/Payload/Carbons.php
@@ -28,7 +28,13 @@ class Carbons extends Payload
                 $displayed = new Displayed;
                 $displayed->handle($message->displayed, $message);
             } elseif (count($jingle_messages = $stanza->xpath('//*[@xmlns="urn:xmpp:jingle-message:0"]')) >= 1) {
-                \Moxl\Xec\Handler::handleNode($jingle_messages[0], $message);
+                $callto = current(explode('/', (string)$message->attributes()->to));
+                if ($callto == \App\User::me()->id || $callto == "") {
+                    // We get carbons for calls other clients make as well as calls other clients receive
+                    // So make sure we only ring when we see a call _to_ us
+                    // Or with no "to", which means from ourselves to ourselves, like another client's <accept>
+                    \Moxl\Xec\Handler::handleNode($jingle_messages[0], $message);
+                }
             }
         }
     }


psycotica0:
Let me know if this breaks a style-guide, or if you'd rather I handle this in a completely different way.

I tried to make the patch very small, but perhaps I made it too small. Let me know!

psycotica0:
I may have noticed an issue here, where when I'm making an outbound call Movim rings thinking I'm calling myself.
I'll test this more later

edhelas:
I didn't forget you :)  Don't worry. I'll take time to review properly this interesting pull request.
Regards.

psycotica0:
Oh, right! That reminds me that I should fix the bug I found and test more myself!

psycotica0:
Ok, I have an implementation question for the maintainer. In the current implementation the assumption is that if I receive a `jingle-propose`, it means I'm being called. If the XMPP server gave me a message, it's because it's meant for me.

But with Carbons that's not true anymore, because I now see the calls that other clients are making also being forwarded to me.

Right now, it just rings, which is bad. There are three places I could fix this, but I want to know what you think about them, because they're slightly different.

## `lib/moxl/src/Xec/Payload/Carbons.php`

I could just put an if-statement in this file I've modified already.
It would look something like:

```php
                $callto = current(explode('/', (string)$message->attributes()->to));
                if ($callto == \App\User::me()->id || $callto == "") {
                    // We get carbons for calls other clients make as well as calls other clients receive
                    // So make sure we only ring when we see a call _to_ us
                    // Or with no "to", which means from ourselves to ourselves, like another client's <accept>
                    \Moxl\Xec\Handler::handleNode($jingle_messages[0], $message);
                }
```

This means that we only process the carbons that contain messages we think we would have been sent normally if there wasn't a higher-priority client running.

This would be the smallest change, because things would mostly work the same way Movim works on master, it would just also ring with Gajim was running, which is the purpose of this PR.

## `lib/moxl/src/Xec/Payload/Jingle*.php`

We could put a very similar version of the above if-statement in these files. So, Carbons gets to be dumb and forward everything on, but here we filter out the ones we don't like.

This is my least favourite option, but I'm including it in case you like it.

## `app/widgets/Visio/Visio.php`

Here is where we receive the events, so we could let the events flow all the way through the system, and then treat them differently here at the listener.

The if-statement would look similar here, but the difference is that we can be more specific if we wanted to.
Even without the original issue (that is, even if we aren't running Gajim), we will receive carbons when another client makes a call. If we filter those out in `Carbons.php`, then it will be just like before. But if we filter it out here we could choose to show that in the chat log, even if we don't ring. If we handle it here, we technically have more information than we used to.

This one may be more work, but I figured I'd mention it in case you feel like this is the right place to handle it.

psycotica0:
BTW, I tested the proposed change to `lib/moxl/src/Xec/Payload/Carbons.php` and it seems to work in the cases I can think of:

- Call that hangs up, Movim stops ringing
- Call that Movim picks up works
- Call that Conversations picks up, Movim stops ringing

Both when Gajim is Open and Closed, and it seems to work the same way in both cases.

I also tested refusing the call in Conversations, but Movim kept ringing. This happened both with and without Gajim running, so I think it's a new bug, actually.

psycotica0:
> I also tested refusing the call in Conversations, but Movim kept ringing. This happened both with and without Gajim running, so I think it's a new bug, actually.

I looked into this, and I think this is a bug in Conversations. They appear to only send the `<reject>` to the caller, but the standard says you should send it to yourself, and *may* send it to the caller. Movim appears to be doing the right thing here, and I think this code would just work if Conversations was doing it right.

**EDIT** It looks like the bug has already been reported (https://github.com/iNPUTmice/Conversations/issues/4000)

singpolyma:
>I looked into this, and I think this is a bug in Conversations. They appear 
>to only send the `<reject>` to the caller, but the standard says you should 
>send it to yourself, and *may* send it to the caller. Movim appears to be 
>doing the right thing here, and I think this code would just work if 
>Conversations was doing it right.

I believe there is an issue on Conversations about this. IIRC Conversations 
currently intentionally violates the spec, believing that outbound carbons 
should handle this and the xep should be amended.

psycotica0:
> I looked into this, and I think this is a bug in Conversations. They appear to only send the `<reject>` to the caller, but the standard says you should send it to yourself, and *may* send it to the caller. Movim appears to be doing the right thing here, and I think this code would just work if Conversations was doing it right.
> I believe there is an issue on Conversations about this. IIRC Conversations currently intentionally violates the spec, believing that outbound carbons should handle this and the xep should be amended.

Yeah. Looking into what they do, if we wanted to be compatible, it looks like we'd need to do the `Visio.php` version. The way they handle it is that if the message is for [a session we already know about](https://github.com/iNPUTmice/Conversations/blob/f92ea5c70b292885b164ffb48829ca4f10cd8b52/src/main/java/eu/siacs/conversations/xmpp/jingle/JingleConnectionManager.java#L235-L252), handle the [rejection message we saw someone else send](https://github.com/iNPUTmice/Conversations/blob/f92ea5c70b292885b164ffb48829ca4f10cd8b52/src/main/java/eu/siacs/conversations/xmpp/jingle/JingleRtpConnection.java#L572-L601). But if we see a message from ourselves for a session we don't know anything about, then [we ignore it](https://github.com/iNPUTmice/Conversations/blob/f92ea5c70b292885b164ffb48829ca4f10cd8b52/src/main/java/eu/siacs/conversations/xmpp/jingle/JingleConnectionManager.java#L268-L271).

If we wanted to do something similar here, we'd need the message to be passed around to the point where some code actually has knowledge of sessions, and that code shouldn't be the handler for Carbons 😅

PR #1028, opened 2021-11-05
Add style parsing for XEP-0393 styling directives.

### LINKED ISSUE #1027: Support Message Styling (XEP-0393)
Hello guys, firstly I'd just like to thank you all for the great work you've been doing on this platform! One improvement that I think would be really great though is the addition of support for XEP-0393 which would allow for Markdown-like message styling and I think would really help make the transition more seamless for normies coming from platforms like Discord. Here's the relevant spec: https://xmpp.org/extensions/xep-0393.html

ChelonianEgghead:
This should resolve #1027. Do note that nested quotes haven't been implemented as I don't fully understand how quotes work in Movim, but this should be good enough for a rudimentary implementation. I also had some trouble getting the regex not to match styling directives contained within blocks or spans of preformatted text, so instead I just overrode their styling in the associated CSS stylesheet. I also removed /code and /me as they should no longer be needed now that we have support for XEP-0393.

### DIFF
--- a/app/widgets/Chat/Chat.php
+++ b/app/widgets/Chat/Chat.php
@@ -1056,6 +1056,12 @@ class Chat extends \Movim\Widget\Base
 
     public function prepareMessage(&$message, $jid = null)
     {
+        $message->body = (preg_replace ('/```((.|\n)*?)(```|\z)/', "<pre>$1</pre>", $message->body));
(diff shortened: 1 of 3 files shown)


edhelas:
Integrated in https://github.com/movim/movim/commit/865ab1ead41ad35713c5a02a94aea84666f2efd3

PR #1040, opened 2022-02-04
Correct XEP-0393 Style Parsing

ChelonianEgghead:
From the XEP:

> Spans are always the children of blocks and may not escape from their containing block. Matches of spans between two styling directives MUST contain some text between the two directives, otherwise neither directive is valid. The opening styling directive MUST be located at the beginning of the parent block, after a whitespace character, or after a different opening styling directive. The opening styling directive MUST NOT be followed by a whitespace character and the closing styling directive MUST NOT be preceeded by a whitespace character. Spans are always parsed from the beginning of the byte stream to the end and are lazily matched. Characters that would be styling directives but do not follow these rules are not considered when matching and thus may be present between two other styling directives.

### DIFF
--- a/app/widgets/Chat/Chat.php
+++ b/app/widgets/Chat/Chat.php
@@ -1101,10 +1101,10 @@ class Chat extends \Movim\Widget\Base
 
         // XEP-0393
         // $message->body = (preg_replace ('/```((.|\n)*?)(```|\z)/', "<pre>$1</pre>", $message->body));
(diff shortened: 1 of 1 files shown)


ltguillaume:
I don't think these expressions are fully correct (although it won't affect the outcome):

`[\s,\*,_,~]` means "a single character, either a whitespace character, a comma, an asterisk, a comma, an underscore, a comma or a tilde.

Essentially it's the same as `[\s\*_,~]`

PR #1058, opened 2022-05-13
DOAP: Enable logo and add "Browser" os/platform

cal0pteryx:
xmpp.org now lists clients by their DOAP files (with image and platform infos), see https://xmpp.org/software/clients/


cal0pteryx:
@edhelas ?

PR #1072, opened 2022-07-21
Add Caddyfile example

Rabioli:
Tested to work with my pod/instance.

May want to add some reference to it in the installation guide and update it, I would make it myself, but I don't know if I can even modify the wiki or how to do so.

### DIFF
--- a/etc/caddy/Caddyfile
+++ b/etc/caddy/Caddyfile
@@ -1,0 +1,53 @@
+your.domain.tld {
+	# You need to change:
+	# Your domain/subdomain
+	# The unix php socket used to match the one of your system
+	# The port you set for movim under /ws/*
+	# The root directory where movim is installed <path-to>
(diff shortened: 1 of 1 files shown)


edhelas:
Im not a big fan of putting the PHP version directly in the config file like this, because it will be quickly outdated.
Regarding the Wiki, feel free to add a new section there based on the two other servers (Apache and nginx) https://github.com/movim/movim/wiki/Install-Movim#4-configure-your-web-server :)

Rabioli:
I already added/modified the entry of the wiki, as for the PHP version I used the one that is on the caddy documentation and also happens to be on debian-stable, it may be a bit outdated, but I think everyone could figure out the version of fpm they are running, I could change it if you want but to what?

edhelas:
For nginx there's an external file that contains all the php-fpm things https://github.com/movim/movim/blob/master/etc/nginx/conf.d/movim.conf#L19

I'd say that it's not the responsibility of the Movim server config file to enforce a specific PHP config.

PR #1111, opened 2022-11-14
Update INSTALL.md

lyn1337:
mention the my-/mariadb collation requirement for case sensitivity


edhelas:
Looks great, can you just add `code quotes` around to make those collations pop out ? And i'll merge :)

edhelas:
:+1:

PR #1112, opened 2022-11-17
Provide a runit service example

lyn1337:
… to be used instead of systemd wherever runit is available (Void Linux, Devuan, antiX, Artix, Venom, Dragora/Linux-Libre)


Rabioli:
I would use /bin/sh for the shell scripts and try to use POSIX so it is more portable and non dependent on bash.

lyn1337:
Fair, dash is generally better for scripts and I think we could just use it instead. On my systems, I don't - and I think those dists have it present if not even as the default shell. Really just an example! Hope more people will consider using runit.

edhelas:
Would it be possible to not duplicated https://github.com/movim/movim/pull/1112/files#diff-8eacd4e0cda0fbbab71b4f13af9602de29d0e3f904909cb516b8a0319f890c40 and simply use an existing .env file for it ? 

Each time I'll add or remove an environment variable I'll have to update this file as well in the file tree.

edhelas:
Bumping on that one.

edhelas:
As I didn't had any feedback regarding this PR, I'm closing it, feel free to reopen one with the addressed changed.

Issue #1046, opened 2022-02-19
Corrupted avatars


TheBluestBird:
Just updated to new Movim 0.20
Avatars look distorted, like here [Blue blog](https://chat.macaw.me/?blog/Blue)
I run Archlinux with php-fpm 7.4.28, people say it might matter that my my php7-imagick is 3.7.0-2

lllvvv:
I have seen the same thing happening, and not only with avatars. Also running Arch, but in my case with php-fpm 8.1.3 and the same version of php-imagick. I can try and help debug this with some instructions, because this bug is breaking the experience and making movim near unusable.

edhelas:
This seems to be caused by php-imagick indeed. I'm keeping this ticket open for a while.
If you find the real issue in php-imagick or an already opened ticket do not hesitate to put the link there.

edhelas:
Hi, any news regarding that issue ?

edhelas:
Closing this ticket as I cannot reproduce

TheBluestBird:
I'm afraid this is something you will face in some time, pretty long knowing the freshness of apps in debian, but still limited.
I really have absolutely no Idea about PHP, I would like to help somehowm but I just have no Idea what and where do I do!

Can you take a look, can this be a solution? Can you please guide me if there is a way to test it?
https://stackoverflow.com/questions/30078090/imagewebp-php-creates-corrupted-webp-files

edhelas:
Actually Movim is now only using Imagick and not GD anymore, so `imagewebp` is not even used in the code.
More investigation will be required. I don't have an easy way to reproduce that issue on my Debian :/

TheBluestBird:
Sorry, my bad, is there any experiment I can conduct to help investigation?

ghost:
Hello

Running Movim on Gentoo, I found after a tedious search that the issue comes from imagemagick version 7.x ( 7.1.0.13 in my case on Gentoo ). After a downgrade to imagemagick 6.9.12.59 my avatar is fine.

TheBluestBird:
This feels like a very strange situation: there is a bug in upstream version of a very popular library little yet know about, and the only way out is to rollback... I still feel like I have just misconfigured something, it can't be this bad!

TheBluestBird:
Interesting, I just have done some experiment, I have just ran command line conversion with `convert xxx.png xxx.webp` and the file happened to be fine. How relevant is it? May be we're blaming ImageMagick for no reason?

[...]

PR #1122, opened 2022-12-10
Bug/corruptedAvatars

TheBluestBird:
I have an [issue](https://github.com/movim/movim/issues/1046) with my installation, it looks like a really distorted images for avatars.
The setup is like 
php 8.1.12 (but was appearing on earlier versions)
imagemagick 7.1.0.53 (but was appearing on earlier versions)
php-imagick 3.7.0 (but was appearing at least since 3.5.1)

This solution helps but I have no idea if it keeps things in order, please check it!

### DIFF
--- a/src/Movim/Image.php
+++ b/src/Movim/Image.php
@@ -208,13 +208,13 @@ class Image
                 $this->_im->setInterlaceScheme(\Imagick::INTERLACE_PLANE);
 
                 // Put 11 as a value for now, see http://php.net/manual/en/imagick.flattenimages.php#116956
-                $this->_im->setImageAlphaChannel(\Imagick::ALPHACHANNEL_REMOVE);
+                $this->_im->setImageAlphaChannel(\Imagick::ALPHACHANNEL_OFF);
                 //$this->_im->setImageAlphaChannel(11);
                 $this->_im->setImageBackgroundColor('#ffffff');
             }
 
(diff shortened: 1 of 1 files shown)


edhelas:
Hi, those constants are only available since Image Magick 7.0. 
I'll have to add a test around to check the version. Can you please drop me there the return of this call:

```
var_dump(Imagick::getVersion());
```

For me I have this:

```
array (size=2)
  'versionNumber' => int 1691
  'versionString' => string 'ImageMagick 6.9.11-60 Q16 x86_64 2021-01-25 https://imagemagick.org' (length=67)
```

TheBluestBird:
Sure, here it is

```
array(2) {
  ["versionNumber"]=>
  int(1808)
  ["versionString"]=>
  string(66) "ImageMagick 7.1.0-53 Q16-HDRI x86_64 20646 https://imagemagick.org"
}
```

edhelas:
I'd like to know what is the versionNumber of ImageMagick 7.0 then, for now I have no clue :/

TheBluestBird:
You want to tell it specifically from int versionNumber?
Yeah, quick browsing didn't help me much, i'm going to look some more little later, maybe be there is some version map somewhere...

Would you consider a dirty hack if we check it from regexing versionString? We could do it once on the launch an save it as a constant, not to do it everytime on conversion

PR #1127, opened 2023-01-01
Support DTMF during audio-only calls

### LINKED ISSUE #1063: Feature: add telephone keypad to Audio Call?
I use the mov.im client to call customer support lines (with jmp.chat).

Some customer support services require the caller (me) to type options on the keypad e.g. "if you want to talk to this department type 1, if you want to talk this other department type 2" or "type # to confirm".

There is no keypad on the pop up window when calling (Audio Call.)

Would adding a telephone keypad to the Audio Call pop up window be a possibility? 

https://en.wikipedia.org/wiki/Keypad

singpolyma:
Fixes #1063

### DIFF
--- a/app/widgets/Visio/visio.css
+++ b/app/widgets/Visio/visio.css
@@ -135,6 +135,12 @@ body {
     opacity: 0;
 }
 
+#visio #dtmf {
+    font-size: 3em;
+    margin-top: calc(50vh - 8em);
+    text-align: center;
+}
+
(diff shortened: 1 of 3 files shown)


edhelas:
Thanks for this awesome pull request, I'll have a look at it :) !

edhelas:
Fixed in https://github.com/movim/movim/commit/f8f30c2307c2937aaa194af1ab0c44816bd4ddce

I did a few changes for better integration ;)

PR #1140, opened 2023-01-22
Movim does not support any commands

singpolyma:
Movim does support fetching commands from others as a client, but does not serve any commands itself, thus it as at best misleading to advertise support for being a command service.


Ppjet6:
LGTM

edhelas:
Indeed ! Thanks :)

PR #1144, opened 2023-01-23
Result and reported forms

### LINKED ISSUE #1093: Support for reported/items elements in adhoc command results
Movim does not display anything when an adhoc command returns this:

```xml
<iq id="Qg9Q5I" type="result" xml:lang="en" from="dummy.localhost" to="test@localhost/movimo9rKMR">
    <command xmlns="http://jabber.org/protocol/commands" node="search" sessionid="xxxx" status="completed">
        <x xmlns="jabber:x:data" type="result">
            <title>Contact search results</title>
            <reported>
                <field var="name" type="text-single" label="Name" />
                <field var="jid" type="jid-single" />
            </reported>
            <item>
                <field var="name">
                    <value>bubu</value>
                </field>
                <field var="jid">
                    <value>bubu@dummy.localhost</value>
                </field>
            </item>
        </x>
    </command>
</iq>
```

Maybe because this uses the `<reported><item>...` syntax and movim does not recognize it?

### LINKED ISSUE #1095: Adhoc command result containing JIDs are not clickable
When the result of an adhoc command contains "jid-single" or "jid-multi" fields,
it would be nice if they were hypertext links.

Example stanza:
```xml
<iq id="3nU8Yq" type="result" xml:lang="en" from="dummy.localhost" to="test@localhost/movimo9rKMR">
    <command xmlns="http://jabber.org/protocol/commands" node="info" sessionid="XXX" status="completed">
        <x xmlns="jabber:x:data" type="result">
            <title>Component info</title>
            <field type="jid-multi" label="Users">
                <value>test@localhost</value>
            </field>
        </x>
    </command>
</iq>
```

Here test@localhost could open the test@localhost chat view.

One possible problem with this is that we don't know if the JID is a MUC, a component, or a "normal" client JID.
Opening the direct message view when clicking the JID sounds reasonable to me.

singpolyma:
Fixes #1095
Fixes #1093

### DIFF
--- a/lib/XMPPtoForm.php
+++ b/lib/XMPPtoForm.php
@@ -3,6 +3,7 @@
 class XMPPtoForm
 {
     private $xmpp;
+    private $formType;
     private $stanza;
     private $html;
 
@@ -15,6 +16,7 @@ class XMPPtoForm
     public function getHTML(\SimpleXMLElement $xmpp, $stanza = false)
     {
         $this->xmpp = $xmpp;
+        $this->formType = $xmpp->attributes()->type;
         $this->stanza = $stanza;
         $this->create();
         return $this->html->saveHTML();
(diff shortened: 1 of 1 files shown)


edhelas:
The rest looks fine :)

singpolyma:
@edhelas fixed

PR #1169, opened 2023-02-28
Fix scrollbar width in Firefox, move scrollbar specifics to scrollbar.css

ltguillaume:
Apparently `scrollbar-width` doesn't propagate in Firefox, so it needs to be set for all elements.

### DIFF
--- a/public/theme/css/color.css
+++ b/public/theme/css/color.css
@@ -8,9 +8,6 @@ body {
     --movim-gray: 35, 35, 35;
     --movim-font: 0, 0, 0;
     --movim-element-action: 153, 153, 153;
-    scrollbar-color: rgba(var(--movim-font), 0.2) transparent;
-    scrollbar-width: thin;
-
     --movim-red: #F44336;
 }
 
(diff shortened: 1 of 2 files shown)


edhelas:
I don't see the difference with your changes. 
Actually `scrollbar-width: thin;` doesn't look nice to me. I'll remove it to just use the default width.

ltguillaume:
> I don't see the difference with your changes.

Does that mean you can see the thin scrollbar in Firefox _without_ this change, or that the change didn't work for you?

edhelas:
I'll remove all the CSS related to scrollbar thickness and just let the browser handle that part.

ltguillaume:
I hope you mean just `scrollbar-width: thin;` for Firefox, because the styling for Chromium looks great and is really necessary.

edhelas:
image

In Firefox `thin` makes the scrollbar barely noticeable. I prefer to use the default width.

ltguillaume:
Wow, that looks odd indeed. Looks like this for me in FF:

image

PR #1175, opened 2023-03-18
Show ad-hoc command list on mobile

singpolyma:
Instead of hiding the sidebar, move it to the bottom.


edhelas:
No, this is really changing a lot the UI. I'd prefer to rethink this and move things in a proper slidable sidebar or something.

PR #1176, opened 2023-03-18
Nice UI for adding a contact through a Gateway

singpolyma:
Uses jabber:iq:gateway where supported (such as slidge and cheogram) and
otherwise falls back to jid escaping where supported, or finally traditional %
escaping.  When searching for contacts, automatically search all gateways that
are added to contacts as well, and show any relevant results.

Note: if the gateway returns an error because the format is bad, this doesn't
show the error since that doesn't really fit the search paradigm, and instead it
simply does not show a result for that gateway.

Also contains a patch to show sms contacts with the "mobile" icon and gateways themselves with the "bot" icon.

### DIFF
--- a/app/Presence.php
+++ b/app/Presence.php
@@ -32,8 +32,7 @@ class Presence extends Model
 
     public function capability()
     {
-        return $this->hasOne('App\Info', 'node', 'node')
-            ->whereNull('server');
+        return $this->hasOne('App\Info', 'node', 'node');
     }
 
     public function contact()
--- a/app/widgets/Search/Search.php
+++ b/app/widgets/Search/Search.php
@@ -5,6 +5,8 @@ use Movim\Widget\Base;
 use Respect\Validation\Validator;
 use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Database\Capsule\Manager as DB;
+use Moxl\Xec\Action\IqGateway;
+use Moxl\Utils;
 
 use App\Post;
 use App\Tag;
@@ -19,6 +21,22 @@ class Search extends Base
     {
         $this->addjs('search.js');
         $this->addcss('search.css');
+        $this->registerEvent('iqgateway_set_handle', 'onIqGatewaySet');
+    }
+
+    public function foundJid($prompt, $jid) {
+        $session = \Movim\Session::start();
+        $jids = $session->get('iqgateway_result_'.$prompt, []);
+        $jids[''.$jid] = 1;
+        $session->set('iqgateway_result_'.$prompt, $jids, 360);
+    }
+
+    public function onIqGatewaySet($packet)
+    {
+        $jid = $packet->content['query']->jid;
+        $prompt = $packet->content['prompt'];
+        $this->foundJid($prompt, $jid);
+        $this->rpc('Search.searchCurrent');
     }
 
     public function ajaxRequest($chatroomActions = false)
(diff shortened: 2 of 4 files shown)


edhelas:
Thanks for the PR ! I'll try to review it in the upcoming days and give you feedback :)

app/Presence.php
```diff
@@ -32,8 +32,7 @@ public function roster()
 
     public function capability()
     {
-        return $this->hasOne('App\Info', 'node', 'node')
-            ->whereNull('server');
```
edhelas: This is risky, I'll have to double check if it doesn't bring some side issues

app/Info.php
```diff
@@ -207,13 +207,15 @@ public function getPhoto($size = 'm')
...
+        }
         if ($this->identities->contains('type', 'handheld')
-        || $this->identities->contains('type', 'phone')) {
+        || $this->identities->contains('type', 'phone')
+        || $this->identities->contains('type', 'sms')) {
             return 'smartphone';
         }
-        if ($this->identities->contains('type', 'bot')) {
-            return 'memory';
-        }
```
edhelas: Would it be possible to move those changes in a separate PR ? I'd prefer to keep things separated.

app/widgets/Search/Search.php
```diff
@@ -19,6 +21,22 @@ public function load()
     {
         $this->addjs('search.js');
         $this->addcss('search.css');
+        $this->registerEvent('iqgateway_set_handle', 'onIqGatewaySet');
+    }
+
+    public function foundJid($prompt, $jid) {
+        $session = \Movim\Session::start();
```
edhelas, line 28: Not a big fan of using memory session for that, I'd prefer to have just parameters passed around.

app/widgets/Search/Search.php
```diff
@@ -140,6 +166,26 @@ public function prepareEmpty()
...
     public function ajaxSearch($key)
     {
+
+        $session = \Movim\Session::start();
+        if ($key && !$session->has('iqgateway_result_'.$key)) {
+            foreach ($this->user->session->contacts()->with('presence.capability')->get() as $contact) {
+                if (!$contact || !$contact->presence || !$contact->presence->capability) continue;
+                if ($contact->presence->capability->hasFeature("jabber:iq:gateway")) {
+                    $r = (new IqGateway\Set)
+                        ->setTo($contact->presence->jid . ($contact->presence->resource ? '/' . $contact->presence->reserouce : ''))
```
edhelas: Typo there

app/widgets/Search/Search.php
```diff
@@ -140,6 +166,26 @@ public function prepareEmpty()
...
+                        ->setPrompt($key)
+                        ->request();
+                } else if ($contact->presence->capability->getGatewayTypeAttribute()) {
+                    if ($contact->presence->capability->hasFeature("jid\\20escaping")) {
+                        $this->foundJid($key, Utils::escapeJidLocalpart($key).'@'.$contact->presence->jid);
```
edhelas, line 181: Rebuilding the JID this way seems risky

edhelas (changes requested):
Is there a way for me to properly test it ?

app/widgets/Search/Search.php
singpolyma: We need to store state somewhere, since this data comes in asynchronously from XMPP and there can be more than one.  The search results are rendered using `draw` and expect the entire list of results to be available for that.  Additionally, we do probably want to cache this for a short time since it's a remote lookup.

I'm not saying memory session is definitely right for it, but I'm curious what you would suggest as an alternative here in order to have `prepareSearch` know about all the items that have come in asynchronously?

app/widgets/Search/Search.php
singpolyma: It's not "rebuilding" but rather building a new JID out of the query and the gateway, having already determined in the line before that this escaping format is supported by the gateway.

We could put the operation in a method somewhere, but it's always going to do fundamentally what this line does.

Neustradamus:
@edhelas: Have you seen @singpolyma changes?

truenicoco:
> Is there a way for me to properly test it ?

Given the age, this PR is probably stale, but slidge does support the protocol this PR is about.

PR #1178, opened 2023-03-19
Use true JID of message sender when we know it

singpolyma:
Save true counterpart JID of MUC sender when we know it.

Set true counterpart JID by XEP-0033 ofrom, but only when it is quite safe, that is when the domain of the ofrom matches the domain of the from, and that domain advertises support for XEP-0033 itself.  (Allowing for fallback body in conjunction with this.)

Show nickname and icon based on true counterpart JID when we know it.

### DIFF
--- a/app/Message.php
+++ b/app/Message.php
@@ -296,9 +296,10 @@ class Message extends Model
             $parentMessage = $this->resolveParentMessage($this->jidfrom, (string)$stanza->reactions->attributes()->id);
 
             if ($parentMessage) {
-                $resource = $this->isMuc()
+                $resource = $this->counterpartjid
+                    ? $this->counterpartjid : ($this->isMuc()
                     ? $this->resource
-                    : $this->jidfrom;
+                    : $this->jidfrom);
 
                 $parentMessage
                     ->reactions()
@@ -585,7 +586,7 @@ class Message extends Model
     public function resolveColor()
     {
         $this->color = stringToColor(
-            $this->resource . $this->type
+            ($this->isMuc() ? $this->resource : ($this->counterpartjid ? $this->counterpartjid : $this->fromjid)) . $this->type
         );
 
         return $this->color;
--- a/app/Message.php
+++ b/app/Message.php
@@ -277,6 +277,17 @@ class Message extends Model
             }
         }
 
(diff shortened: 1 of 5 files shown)


Ppjet6:
In which context does Movim do 0033? (TIL).

I would suggest not to use the realjid when we're not `whois = anyone` and we're a moderator. I don't know if it's comparable but 392 does this, bases the color on realjids, and it's actually annoying seeing colors change for the same person in different rooms just because we have different permissions.

Also, not much we can do as long as there's no way to do validation, but this is probably going to come bite us at some point: `$presence->mucjid != $this->jidfrom.'/'.$this->resource`

edhelas:
I still don't know exactly why this PR is for, I'll try to figure it out :thinking:

singpolyma:
>In which context does Movim do 0033? (TIL).

In this context that I just wrote in this PR.

>I would suggest not to use the realjid when we're not `whois = anyone` and we're a moderator. I don't know if it's comparable but 392 does this, bases the color on realjids, and it's actually annoying seeing colors change for the same person in different rooms just because we have different permissions.

This PR does not use realjid for colour in MUC.  It does use it to show both 
your name for a person as well as their MUC resource, and get easy access to 
their JID right on the message when it is known.

-- 
Stephen Paul Weber, @singpolyma
See <http://singpolyma.net> for how I prefer to be contacted
edition right joseph

singpolyma:
>I still don't know exactly why this PR is for, I'll try to figure it out 
>:thinking:

For me the biggest purpose is the 0033 implementation I added, but being 
able to see my name for a MUC user to know who they are is also very 
helpful, etc.

-- 
Stephen Paul Weber, @singpolyma
See <http://singpolyma.net> for how I prefer to be contacted
edition right joseph

singpolyma:
Just following up here with respect to the "more info" label -- happy to provide more info if needed :)

ItsRogueRen:
Just adding my comments here as well, I have chats where I'd need to see my local name for a contact so I can tell who is who in a group of say coworkers who all have weird usernames

nabzord:
+1 merge this

Neustradamus:
Any news on this PR?

PR #1180, opened 2023-03-20
Show alias on profile as well

singpolyma:
So you know who you're looking at.


edhelas:
Not this way, because truename can also contain the alias, so you'll have sometimes twice the same string appearing.
Also this should appear properly the same way a bit everywhere and not only in this card.

Leaving this thing open to see if we can figure a way to handle it nicely.

singpolyma:
This does check already if the roster name and the truename match to make sure you don't show them duplicated.

PR #1181, opened 2023-03-21
Show IMPP URL if there is one

singpolyma:
Allows a contact to list an alternate chat address or JabberID at which they can be reached.

### DIFF
--- a/app/Contact.php
+++ b/app/Contact.php
@@ -230,6 +230,10 @@ class Contact extends Model
             ? (string)$vcard->url->uri
             : null;
 
+        if ($this->url == null && $vcard->impp->uri) {
+            $this->url = $vcard->impp->uri;
+        }
+
         $this->adrlocality = !empty($vcard->adr->locality)
             ? (string)$vcard->adr->locality
             : null;
--- a/app/widgets/ContactData/_contactdata_card.tpl
+++ b/app/widgets/ContactData/_contactdata_card.tpl
@@ -64,7 +64,10 @@
 
                 {if="$contact->url != null"}
                     <i class="material-icons icon-text">link</i>
-                    {if="filter_var($contact->url, FILTER_VALIDATE_URL)"}
+                    {if="parse_url($contact->url, PHP_URL_SCHEME) == 'xmpp'"}
+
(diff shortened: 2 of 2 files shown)


app/widgets/ContactData/_contactdata_card.tpl
```diff
@@ -64,7 +64,10 @@
 
                 {if="$contact->url != null"}
                     <i class="material-icons icon-text">link</i>
-                    {if="filter_var($contact->url, FILTER_VALIDATE_URL)"}
+                    {if="parse_url($contact->url, PHP_URL_SCHEME) == 'xmpp'"}
+
+                        <a href="{$contact->url}" onclick="MovimUtils.reload('{$c->route('contact', substr($contact->url, 5))}'); return false" class="u-impp" rel="me">{$contact->url}</a>
```
edhelas: Why the "u-impp" class and `return: false` ?

app/Contact.php
```diff
@@ -229,6 +229,9 @@ public function setVcard4($vcard)
         $this->url = !empty($vcard->url->uri)
             ? (string)$vcard->url->uri
             : null;
+        if (!$this->url && $vcard->impp->uri) {
```
edhelas: I tend to add a carriage return before if() for readability.
Also testing `$this->url == null` seems cleaner :)

singpolyma:
Somebody claiming to be Jaussoin Timothée wrote:
>@edhelas commented on this pull request.
>
>
>
>> @@ -64,7 +64,10 @@
>
>                 {if="$contact->url != null"}
>                     <i class="material-icons icon-text">link</i>
>-                    {if="filter_var($contact->url, FILTER_VALIDATE_URL)"}
>+                    {if="parse_url($contact->url, PHP_URL_SCHEME) == 'xmpp'"}
>+
>+                        <a href="{$contact->url}" onclick="MovimUtils.reload('{$c->route('contact', substr($contact->url, 5))}'); return false" class="u-impp" rel="me">{$contact->url}</a>
>
>Why the "u-impp" class and `return: false` ?

u-impp is just from h-card, it can be removed if not wanted.

return false is to make it use the soft reload instead of following the link 
after

PR #1185, opened 2023-03-27
Show gateways like bots and sms items as mobile

singpolyma:


app/Presence.php
```diff
@@ -32,8 +32,7 @@ public function roster()
 
     public function capability()
     {
-        return $this->hasOne('App\Info', 'node', 'node')
```
edhelas: I have to check about that change. Wondering if the constraint was not required, do you really need to remove the "where" there ? 

There is no index on that unique column, only on both:

    "infos_server_node_unique" UNIQUE CONSTRAINT, btree (server, node)

singpolyma: In my testing it seems that Info from a domain JID is stored with a non-null `server` column which is why this got removed, otherwise such a contact just never has any capability.

PR #1191, opened 2023-03-31
Fix #1183  compileOpcache fails with php8.1 

### LINKED ISSUE #1183: compileOpcache fails with php8.1
Hi, I installed the latest movim from the master branch as explained [here](https://github.com/movim/movim/blob/master/INSTALL.md), my PHP Version is:
```
www-data@machine:/usr/share/movim$ php -v
PHP 8.1.17 (cli) (built: Mar 16 2023 14:37:38) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.17, Copyright (c) Zend Technologies
    with Zend OPcache v8.1.17, Copyright (c), by Zend Technologies
```
However somehow the compileOpcache is not running properly. I always get the following [error log](https://termbin.com/uxon) when starting the damon.php. I found out that the error reporting is [disabled](https://github.com/movim/movim/blob/master/app/helpers/UtilsHelper.php#L124) for the opcache compile process. 
To get the Error I changed the function temporarely from:
```
/**
 * Compile main files in Opcache
 */
function compileOpcache()
{
    error_reporting(0);
    foreach (listOpcacheCompilableFiles() as $file) {
        opcache_invalidate($file, true);
        yield @opcache_compile_file($file);
    }
    error_reporting(1);
}
```
to 
```
/**
 * Compile main files in Opcache
 */
function compileOpcache()
{
//    error_reporting(0);
    foreach (listOpcacheCompilableFiles() as $file) {
        opcache_invalidate($file, true);
//        yield @opcache_compile_file($file);
        opcache_compile_file($file);
    }
//    error_reporting(1);
}

```
and did a manual compile via `php daemon.php compileOpcache`. This is the output: https://termbin.com/1e4e

ravermeister:
According to https://github.com/movim/movim/issues/1183#issuecomment-1486737189

### DIFF
--- a/app/helpers/UtilsHelper.php
+++ b/app/helpers/UtilsHelper.php
@@ -123,8 +123,11 @@ function compileOpcache()
 {
     error_reporting(0);
     foreach (listOpcacheCompilableFiles() as $file) {
-        opcache_invalidate($file, true);
-        yield @opcache_compile_file($file);
+        if (opcache_is_script_cached($file)) {
+            yield @opcache_invalidate($file, true);
+        } else {
+            yield @opcache_compile_file($file);
+        }
     }
     error_reporting(1);
 }


app/helpers/UtilsHelper.php
```diff
@@ -123,8 +123,11 @@ function compileOpcache()
 {
     error_reporting(0);
     foreach (listOpcacheCompilableFiles() as $file) {
-        opcache_invalidate($file, true);
-        yield @opcache_compile_file($file);
+        if (opcache_is_script_cached($file)) {
+            yield @opcache_invalidate($file, true);
```
edhelas, line 127: There we should then compile just after no ?

ravermeister: No, because a force invalidate lead's to a recompile: https://www.php.net/manual/en/function.opcache-invalidate.php#119026

edhelas: Ok :)  !

PR #1215, opened 2023-06-04
fix apache rewrite rule

ravermeister:
fix apache rewrite rule to reflect the current nginx rule:

        if (!-e $request_filename)
        {
            rewrite ^/(.*) /index.php?query=$1 last;
        }


edhelas:
Awesome, thanks !

PR #1280, opened 2024-03-12
Respect user agent font sizing

toastal:
While the typical user agent font size is `16px`, this value is configurable by the user. Much of this project is built in `rem` unit which are relative to the base which can be helpful to accessibility. However, when a base size is set to `px` the user agent’s font size is no longer respected. This value is a percentage to match the previous `px` value (i.e. 8.25 ÷ 16).

Prior arts typically use 62.5% as it makes `rem`s effectively ⅒ the `px` size which makes the math of converting much easier. Rather than touching all `rem` values, this fix merely is changing the single base value.

[^1]

### DIFF
--- a/public/theme/css/style.css
+++ b/public/theme/css/style.css
@@ -16,7 +16,7 @@
 }
 
 html {
-    font-size: 8.25px;
+    font-size: 51.5625%;
 }
 
 body {


edhelas:
Do you actually have a source that says that using px in `<html>` tend to break the user agent configuration ? Browsers are doing conversion between all the time and I see lots of tutorials that explains to set the base value in px (16px by default).

toastal:
My eyes are not great as they used to be so I noticed immediately as I have the base font size bumped a hair. This one is easy to test just setting the HTML as a `%` as this patch suggests & adjust the user agent font size such as `about:preferences#fontsGroup` in a Gecko-based browser. Trying not to appeal to authority, but I have been doing front-end dev for over a decade.

Seeming genesis of 62.5% body: https://snook.ca/archives/html_and_css/font-size-with-rem

https://www.freecodecamp.org/news/override-root-font-size-for-a-better-user-experience/

https://medium.com/@jagadishkamuni/respecting-font-size-preferences-rems-and-62-5-base-font-size-aleksandr-hovhannisyan-f0473d6c8410

PR #1281, opened 2024-03-12
Load widgets from absolute path

toastal:
If the public cache is located or symlinked elsewhere, the relative path lookup fails. In the case of NixOS, stateful storage (e.g. caches) must be located outside of the stateless store where Movim application code lives.

[^1]


edhelas:
Awesome thanks !

PR #1282, opened 2024-03-12
Allow scheme-less base URIs to pass the validator

toastal:
Scheme isn’t required & instances can operate without. This helps those doing testing or those that wish to support both HTTP & HTTPS.

[^1]

### DIFF
--- a/src/Movim/Console/DaemonCommand.php
+++ b/src/Movim/Console/DaemonCommand.php
@@ -57,8 +57,21 @@ class DaemonCommand extends Command
 
         $loop = Loop::get();
 
-        if (config('daemon.url') && Validator::url()->notEmpty()->validate(config('daemon.url'))) {
-            $baseuri = rtrim(config('daemon.url'), '/') . '/';
+        if (config('daemon.url')) {
+            # Underlying `Validator`, URL validation is done via
+            # FILTER_VALIDATE_URL, where the FILTER_FLAG_SCHEME_REQUIRED was
+            # deprecated & then removed in later PHP versions now *reqiring* a
+            # scheme. Base URIs are allowed to be scheme-less which helps users
+            # looking to support both HTTP & HTTPS (either for testing purposes
+            # or otherwise). As such, we can append a missing scheme in order
+            # satisfy this validator requirement, but still use the scheme-less
+            # URI in practice.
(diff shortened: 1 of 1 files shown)


src/Movim/Console/DaemonCommand.php
```diff
@@ -57,8 +57,21 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        if (config('daemon.url')) {
+            # Underlying `Validator`, URL validation is done via
+            # FILTER_VALIDATE_URL, where the FILTER_FLAG_SCHEME_REQUIRED was
+            # deprecated & then removed in later PHP versions now *reqiring* a
+            # scheme. Base URIs are allowed to be scheme-less which helps users
+            # looking to support both HTTP & HTTPS (either for testing purposes
+            # or otherwise). As such, we can append a missing scheme in order
+            # satisfy this validator requirement, but still use the scheme-less
+            # URI in practice.
+            $scheme_enforced_url = parse_url(config('daemon.url'), PHP_URL_SCHEME)
```
edhelas: I am using camelCase everywhere in Movim :)

src/Movim/Console/DaemonCommand.php
```diff
@@ -57,8 +57,21 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                ? config('daemon.url')
+                : 'http://' . ltrim(config('daemon.url'), '/');
```
edhelas, line 71: Is it wise to actually put HTTP there ?

src/Movim/Console/DaemonCommand.php
toastal: https://github.com/movim/movim/blob/master/src/Movim/Controller/Base.php#L16
https://github.com/movim/movim/blob/master/src/Movim/Controller/Base.php#L17
https://github.com/movim/movim/blob/master/src/Movim/Controller/Base.php#L20

src/Movim/Console/DaemonCommand.php
toastal: Did you miss the comment? It’s just a temp variable to pass the validator for ‘valid’ URLs. Scheme-less URLs work otherwise, it’s just this validator causing system to choke.

edhelas:
Thanks !

toastal:
The big, fat comment was to help explain why a hacky solution would be needed.
-- 

PR #1286, opened 2024-03-14
Add missing CLI config options

toastal:
Only some of these values were configurable via CLI. While it seems most options have migrated to the Admin panel, there is still value in allowing these options to be easily configurable without logging in. Use case: NixOS configurations tend to have as many properties as possible in the declarative config rather than using mutable state; as server configurations tend to be configured once then left alone, this feel like the right UX.

Omitted: banner stuff as it isn’t as simple as a line of configuration needing to access the network, process images, upload, etc.

[^1]

### DIFF
--- a/app/Configuration.php
+++ b/app/Configuration.php
@@ -11,6 +11,7 @@ class Configuration extends Model
     private static $instance = null;
 
     public $fillable = [
+        'chatonly',
         'description',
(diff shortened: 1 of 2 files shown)


edhelas:
Looks good to me :) Resolve the conflict and I can merge, thanks !

PR #1290, opened 2024-03-16
Add <html lang>

toastal:
Important for screen readers, also helps in some spell checking & hyphenation scenarios
https://dequeuniversity.com/rules/axe/4.4/html-has-lang

[^1]

### DIFF
--- a/app/views/page.tpl
+++ b/app/views/page.tpl
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html translate="no">
+<html translate="no" lang="<%language%>">
     <head>
         <meta charset="utf-8" />
         <title><%title%></title>
--- a/src/Movim/Template/Builder.php
+++ b/src/Movim/Template/Builder.php
@@ -18,6 +18,7 @@ class Builder
     private $commonContent = '';
     private $css = [];
     private $scripts = [];
+    private $language = 'en';
     private $dir = 'ltr';
     private $public;
     private $user;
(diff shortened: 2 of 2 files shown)


edhelas:
Is the injected language actually in the correct HTML format ? Thanks for the PR !

toastal:
It appears to work for me getting both the app config & cat swap per user, but I can’t say it is easy for me to get my bearings around OO projects so it could be inefficient (but hopefully not wrong 🤞).
-- 

edhelas:
This should work, however Movim is having a few languages that doesn't respect the HTML standard (https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes)  such as zh_tw or ur_pk (see the list there https://github.com/movim/movim/tree/master/locales) 

Maybe it would be wise to only take the two first letters ?

toastal:
Tough call, but I don’t think you want to lose the region code, but should prefer it when possible (i.e. I would open up an `en-US` translation if it were missing as I much prefer -ize, -or 😆). My preference would either be a second lookup table from current codes to ISO codes *or* rename the files to match the ISO standard (where a `s/_/-/` would be fine).

I would be curious to know (and not surprised) if the browser implicitly falls back to the first two when the region isn’t matched anyhow. In which case, so string manipulation on the locale can work until a better fix is in place.
-- 

toastal:
Actually I may take on this exploding of a string now.

PR #1297, opened 2024-03-23
Unauthed `dir` + language refactor

toastal:
I didn’t realize that `dir` was missing & lang wasn’t as robust as it should be--& neither worked for unauthenticated users. And speaking of not being robust, the Regex for language parsing has a bug in where if a user sends a script (such as [Serbian](https://en.wikipedia.org/wiki/Serbian_language) that uses both Cyrillic & Latin, with code such as `sr-Latn-RS`), we can get a bug. That code wasn’t the easiest to read so I did a refactor on the block to use named captures as I think it’s easier to read/follow.

This will likely be easiest to read by commit.

[^1]

### DIFF
--- a/src/Movim/Template/Builder.php
+++ b/src/Movim/Template/Builder.php
@@ -140,10 +140,14 @@ class Builder
      */
     public function language(): string
     {
-        $lang = $this->user && $this->user->language
-            ? $this->user->language
-            : Configuration::get()->locale;
-        return Locale::printISO639($lang);
+        if ($this?->user?->language != null) {
+            return Locale::printISO639($this->user->language);
+        }
+
+        $locale = Locale::start();
+        $locale->detect(getenv('language'));
+
+        return Locale::printISO639($locale->language ?? Configuration::get()->locale);
     }
 
     /**
--- a/src/Movim/Template/Builder.php
+++ b/src/Movim/Template/Builder.php
@@ -21,6 +21,7 @@ class Builder
     private array $css = [];
     private array $scripts = [];
     private string $eagerScripts = "/\/(movim_rpc|movim_utils)/";
+    private string $lang = Locale::DEFAULT_LANGUAGE;
     private string $dir = 'ltr';
     private bool $public;
     private ?User $user = null;
@@ -141,13 +142,16 @@ class Builder
     public function language(): string
     {
         if ($this?->user?->language != null) {
-            return Locale::printISO639($this->user->language);
+            $this->lang = $this->user->language;
+            return Locale::printISO639($this->lang);
         }
 
         $locale = Locale::start();
         $locale->detect(getenv('language'));
 
-        return Locale::printISO639($locale->language ?? Configuration::get()->locale);
+        $this->lang = Locale::printISO639($locale->language ?? Configuration::get()->locale);
+
+        return $this->lang;
     }
 
     /**
--- a/src/Movim/Template/Builder.php
+++ b/src/Movim/Template/Builder.php
@@ -9,6 +9,7 @@ use App\Configuration;
 use App\User;
 use Movim\Controller\Ajax;
 use Movim\Widget\Wrapper;
+use Movim\i18n\Dir;
 use Movim\i18n\Locale;
 use stdClass;
 
@@ -22,7 +23,7 @@ class Builder
     private array $scripts = [];
     private string $eagerScripts = "/\/(movim_rpc|movim_utils)/";
     private string $lang = Locale::DEFAULT_LANGUAGE;
-    private string $dir = 'ltr';
+    private Dir $dir = Locale::DEFAULT_DIRECTION;
     private bool $public;
     private ?User $user = null;
     private $jsCheck = true;
@@ -78,7 +79,7 @@ class Builder
 
         $outp = str_replace(
             ['<%scripts%>', '<%meta%>', '<%content%>', '<%common%>', '<%title%>', '<%language%>', '<%dir%>'],
-            [$scripts, $this->meta(), $this->content, $this->commonContent, $this->title(), $this->language(), $this->dir()],
+            [$scripts, $this->meta(), $this->content, $this->commonContent, $this->title(), $this->language(), $this->dir()->value],
             $outp
         );
 
@@ -157,13 +158,13 @@ class Builder
     /**
      * Displays the current font direction
      */
-    public function dir()
+    public function dir(): Dir
     {
         if (isLogged()) {
             $lang = \App\User::me()->language;
 
             if (in_array($lang, ['ar', 'he', 'fa'])) {
-                $this->dir = 'rtl';
+                $this->dir = \Movim\i18n\Locale\Dir::RTL;
             }
         }
 
--- a/src/Movim/Template/Builder.php
+++ b/src/Movim/Template/Builder.php
@@ -160,14 +160,7 @@ class Builder
      */
     public function dir(): Dir
     {
-        if (isLogged()) {
-            $lang = \App\User::me()->language;
-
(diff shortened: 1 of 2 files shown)


src/Movim/Template/Builder.php
```diff
@@ -19,7 +19,8 @@ class Builder
     private string $commonContent = '';
     private array $css = [];
     private array $scripts = [];
-    private string $dir = 'ltr';
+    private string $lang = 'en';
+    private \Movim\i18n\Dir $dir = \Movim\i18n\Dir::LTR;
```
edhelas: I think you can import those namespaces in the header

src/Movim/Template/Builder.php
```diff
@@ -138,55 +139,31 @@ public function title(): string
...
     {
-        $lang = $this->user && $this->user->language
-            ? $this->user->language
-            : Configuration::get()->locale;
-
-        if (empty($lang)) {
-            return 'en';
+        if ($this?->user?->language != null) {
+            $this->lang = $this->user->language;
+            return \Movim\i18n\Locale::printISO639($this->lang);
```
edhelas: Same there, its useful when it's used only once, but if the Class is used several times better import it once.

src/Movim/Template/Builder.php
```diff
@@ -138,55 +139,31 @@ public function title(): string
...
         }
 
-        // String casing for the web
-        $split = array_slice(preg_split('/[_-]/', $lang), 0, 3);
-        $lc = array_shift($split);
-        $langCode = strtolower($lc);
+        $l = \Movim\i18n\Locale::start();
```
edhelas: I'm trying to use more descriptive variable names, you can name it $locale there :)

src/Movim/Template/Builder.php
```diff
@@ -138,55 +139,31 @@ public function title(): string
...
-            }
-        }, $split);
-
-        array_unshift($tags, $langCode);
-
-        return implode('-', $tags);
+        $this->lang = \Movim\i18n\Locale::printISO639(
+            $l?->language != null ? $l->language : Configuration::get()->locale,
+        );
+        return $this->lang;
```
edhelas, line 155: Simple syntax, but I like to put carriage return between block of code and before things like return.

src/Movim/Template/Builder.php
```diff
@@ -138,55 +139,31 @@ public function title(): string
...
     {
-        if (isLogged()) {
-            $lang = \App\User::me()->language;
-
-            if (in_array($lang, ['ar', 'he', 'fa'])) {
-                $this->dir = 'rtl';
-            }
-        }
-
+        $this->dir = \Movim\i18n\Locale::getDirection($this->lang);
```
edhelas, line 163: Class namespace import in the header and carriage return after :)

src/Movim/Template/Builder.php
```diff
@@ -138,55 +139,31 @@ public function title(): string
...
-        $split = array_slice(preg_split('/[_-]/', $lang), 0, 3);
-        $lc = array_shift($split);
-        $langCode = strtolower($lc);
+        $l = \Movim\i18n\Locale::start();
 
-        if (empty($split)) {
-           return $langCode;
+        if (getenv('language') != null) {
+            $l->detect(getenv('language'));
+        } elseif (isset($_SERVER['HTTP_ACCEPT_LANGUAGE'])) {
```
edhelas: It seems that `isset($_SERVER['HTTP_ACCEPT_LANGUAGE'])` is actually called inside `detect()` directly, maybe it can be removed there then and used when the parameter is null

src/Movim/i18n/Locale.php
```diff
@@ -179,43 +187,81 @@ public function translate(string $key, $args = false): string
     /**
      * @desc Auto-detects the language from the user browser
      */
-    public function detect($accepted = false)
+    public function detect($accepted = false): ?string
```
edhelas: Small refactor, we can make this parameter strict and call it `?string $language` that fall back to `HTTP_ACCEPT_LANGUAGE` if the value is null, will be way cleaner.

src/Movim/i18n/Locale.php
```diff
@@ -179,43 +187,81 @@ public function translate(string $key, $args = false): string
...
+                if (isset($loc['region'])) {
+                    unset($loc['region']);
+                    [$lang, $exists] = $poFileExists($loc);
+                    if ($exists) {
+                        $this->language = $lang;
+                        break;
+                    }
+                }
 
+                $this->language = 'en';
```
edhelas: Maybe you can just initiate the variable to `en` at the beginning, it will also prevent the next else condition.

src/Movim/i18n/Locale.php
```diff
@@ -298,4 +344,42 @@ private function getQuotedString(string $string)
...
+    /**
+     * @desc Determine the direction of a locale string
+     */
+    public static function getDirection(string $str): Dir
+    {
+        $loc = locale_parse($str);
+
+        if (isset($loc['script'])) {
+            return in_array($loc['script'], self::$rtlScripts) ? Dir::RTL : Dir::LTR;
+        } else {
```
edhelas: No need of the else there

toastal: Personally I like the symmetry but 🤷

edhelas:
locale_parse() requires the intl library that is not shipped with PHP and quite heavy in RAM.

https://www.php.net/manual/en/locale.parselocale.php

We'll have to do without it.

edhelas:
Can you please rebase and squash a maximum of commits

edhelas:
Could you also do another PR with all the file "last-line carriage return" fixes, those are not related to that one, it's difficult to review.

edhelas:
```
[2024-03-28T11:42:34.966935+00:00] movim.ERROR: TypeError: array_values(): Argument #1 ($array) must be of type array, null given in /var/www/html/movim/src/Movim/i18n/Locale.php (line 383)
Trace
#0 /var/www/html/movim/src/Movim/i18n/Locale.php(383): array_values()
#1 /var/www/html/movim/src/Movim/i18n/Locale.php(391): Movim\i18n\Locale::printPOSIX()
#2 /var/www/html/movim/src/Movim/i18n/Locale.php(281): Movim\i18n\Locale::printPo()
#3 /var/www/html/movim/src/Movim/i18n/Locale.php(73): Movim\i18n\Locale->load()
#4 /var/www/html/movim/src/Movim/Console/CompileLanguages.php(30): Movim\i18n\Locale->compilePos()
#5 /var/www/html/movim/vendor/symfony/console/Command/Command.php(326): Movim\Console\CompileLanguages->execute()
#6 /var/www/html/movim/vendor/symfony/console/Application.php(1078): Symfony\Component\Console\Command\Command->run()
#7 /var/www/html/movim/vendor/symfony/console/Application.php(324): Symfony\Component\Console\Application->doRunCommand()
#8 /var/www/html/movim/vendor/symfony/console/Application.php(175): Symfony\Component\Console\Application->doRun()
#9 /var/www/html/movim/daemon.php(21): Symfony\Component\Console\Application->run()
```

edhelas:
I put the branch as WIP as it seems that several things were not tested properly, once you're done please remove the 'WIP' in the title

toastal:
@edhelas Yes, I had accidentally ran a cached version late last night when I ‘tested’ it & saw it running correctly. I was in the middle of fixing it when you caught it. Currently I ran uncached with Brazilian Portuguese & Hebrew being the import test languages (+ region, RTL).

toastal:
One sec. I see I have some things indented not the same as other (EditorConfig couldn’t cover PHP)

PR #1300, opened 2024-03-25
ImageSize enum

toastal:
Prevent bugs related to unsupported sizes.

[^1]

### DIFF
--- a/src/Movim/ImageSize.php
+++ b/src/Movim/ImageSize.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Movim;
+
+enum ImageSize: string
+{
+    case XXL = 'xxl';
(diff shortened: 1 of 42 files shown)


src/Movim/ImageSize.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Movim;
+
+enum ImageSize: string
```
edhelas, line 5: Does it actually require a specific file for that ? 

Having the enum as a static parameter in Movim/Image should do it no ?

toastal:
I could not figure out how to get linker to autoload it otherwise without missing class errors. I don’t know why this worked & I pray that I never learn enough PHP to truly know why
-- 

Issue #1327, opened 2024-06-21
Firefox Desktop Clients Disconnect When Attempting a Call


Metalhearf:
## Description

When two Firefox clients try to initiate a call, both clients get disconnected from ejabberd and are redirected to the Movim login page, hinting to a probable PHP crash.

## Environment

- Movim: 2068488c579c3cbfcc281ab2a4ebabed689041eb (Thu Jun 13 14:07:47 2024 +0200)
- ejabberd: 24.04

### Movim error.log

```log
[2024-06-18T08:42:47.984835+00:00] movim.ERROR: Invalid characters passed for attempted conversion, these have been ignored in /var/www/movim/src/Movim/Librairies/JingletoSDP.php (line 47)
  
[2024-06-18T08:42:47.985181+00:00] movim.ERROR: TypeError: Illegal offset type in /var/www/movim/src/Movim/Librairies/JingletoSDP.php (line 136)
Trace
#0 /var/www/movim/app/widgets/Visio/Visio.php(121): Movim\Librairies\JingletoSDP->generate()
#1 /var/www/movim/src/Movim/Widget/Wrapper.php(195): Visio->onInitiateSDP(Array)
#2 /var/www/movim/src/Moxl/Xec/Payload/Payload.php(78): Movim\Widget\Wrapper->iterate('jingle_sessioni...', Array)
#3 /var/www/movim/src/Moxl/Xec/Payload/Jingle.php(34): Moxl\Xec\Payload\Payload->event('jingle_sessioni...', Array)
#4 /var/www/movim/src/Moxl/Xec/Handler.php(207): Moxl\Xec\Payload\Jingle->handle(Object(SimpleXMLElement), Object(SimpleXMLElement))
#5 /var/www/movim/src/Moxl/Xec/Handler.php(111): Moxl\Xec\Handler::searchPayload('d84d4b89d43e88a...', Object(SimpleXMLElement), Object(SimpleXMLElement))
#6 /var/www/movim/src/Moxl/Xec/Handler.php(75): Moxl\Xec\Handler::handleNode(Object(SimpleXMLElement), Object(SimpleXMLElement))
#7 /var/www/movim/linker.php(49): Moxl\Xec\Handler::handle(Object(SimpleXMLElement))
#8 [internal function]: {closure}(Object(SimpleXMLElement))
#9 /var/www/movim/src/Moxl/Parser.php(110): call_user_func_array(Object(Closure), Array)
[...]

PR #1328, opened 2024-06-21
Use string casting for senders attribute

Metalhearf:
Potential fix for movim/movim#1327.

Tested: voice and video calls are now available again for different Firefox users.


edhelas:
Thanks ! :)

PR #1341, opened 2024-07-19
Add support of custom inline emojis through XEP-0231: Bits of Binary

edhelas:
- Refactor the stickers and custom emojis by moving them in the database
- Add an importEmojisPack command to import Mastodon/Pleroma emojis pack

### DIFF
--- a/app/EmojisPack.php
+++ b/app/EmojisPack.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace App;
+
+use Illuminate\Database\Eloquent\Model;
+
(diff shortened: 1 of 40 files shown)


app/EmojisPack.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace App;
+
+use Illuminate\Database\Eloquent\Model;
+
+class EmojisPack extends Model
+{
```
truenicoco, line 8: Braces!? Why!? What's next? Semi-colons maybe?

PR #1368, opened 2024-10-11
Multiparty Jingle

edhelas:
- Add support of XEP-0482: Call Invites
- Add support of XEP-0272: Multiparty Jingle
- Create Muji tables in the database
- Add Moxl Muji and Call Invites classes
- Refactoring to allow multiple Jingles at the same time
- Refactor the Jingle XMPP payload to use Packets
- Generalize the from parameter in all the Jingle events
- Create/delete dynamicaly the remote audio and video tags per Jingle session
- Initiate individual Jingle sessions in a MujiCall
- Refactor the presence handling to initiate properly the individual Jingle sessions
- Add adjustments to fit with Dino
- Move the audio/video toggle and screen sharing features to MovimJingles
- Handle properly Call Invite retractation
- Handle state where there is several Muji calls invites at the same time
- Add Ongoing Call in the title bar
- Handle mute/unmute in Muji calls 
- Create new status messages to track the calls in the MUCs
- Complete DOAP


kousu:
This is truly incredible. I never thought I would see the day XMPP caught up to Zoom or Hangouts but here we are. And you've been so modest on this thread about it...

PR #1431, opened 2025-04-23
ci: add mariadb setup job

kskarthik:
- setup jobs run on latest maria & postgres versions
- setup workflows now runs on all branches
- add my info to contributor list


edhelas:
Thanks a lot ! ✨

PR #1532, opened 2026-02-28
Set versions and use full path to images

oappiah:
Make the "Quick Test" https://github.com/movim/movim#quick-test use verified and working versions.
Using full path to registries, default configuration of podman and/or different distributions does not default to docker.io.


poVoq:
I am a bit surprised that it works with Postgres 18. Were there not some breaking changes on that version?

PR #1539, opened 2026-03-13
Add icon for attachment and handling different file formats

divyaranjan1905:
This is a very basic addition to handle different file formats when you upload them. It creates a generic icon to make them look distinct instead of the URL-based look.

**Note:** I haven't properly tested this yet due to conflicts in my test environment of Movim. But if you can test this out and let me know, that'd be great.

### DIFF
--- a/app/Widgets/Chat/chat.css
+++ b/app/Widgets/Chat/chat.css
@@ -803,10 +803,12 @@ ul.list li>div.bubble.file div.file[data-type="image/gif"]:after {
     pointer-events: none;
 }
 
-ul.list li>div.bubble.file div.file a span.host:after,
-ul.list li>div.bubble.file div.file a span.size:after {
-    content: '-';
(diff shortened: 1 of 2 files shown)


divyaranjan1905:
I (think) fixed my container issue, and here's what the attachments look like:
image

divyaranjan1905:
I think there's also an issue with pictures sometimes not previewing them. The picture is clearly there, if you click on it. I'll try to see if I can fix that in this PR too.

divyaranjan1905:
I'm going with material-symbols since they seem to be readily available, but do you have any better suggestions?

divyaranjan1905:
Also, this works only if you're sending the file from/to Movim. Sending, for e.g., from Dino just gives you a link. I think we should handle that as well.

divyaranjan1905:
I'lll create another PR with PHP-based approach.

PR #1544, opened 2026-03-14
Feat: Add attachment icons and MIME-type handling

divyaranjan1905:
This PR adds icons to messages with file attachments, based on their MIME type. This is done primarily on the PHP side instead of JS which was the approach taken in #1539. 

This is how it looks right now:

image

### DIFF
--- a/app/Helpers/UtilsHelper.php
+++ b/app/Helpers/UtilsHelper.php
@@ -1090,3 +1090,58 @@ function base64ToFingerPrint(string $base64): string
     $hex = unpack('H*', $buffer);
     return implode(' ', str_split(substr($hex[1], 2), 8));
 }
+
+/**
+ * @desc Return a Material Symbols icon name for a given mimetype
+ */
+function mimeToIcon(string $type): string
+{
+    return match(true) {
+        $type === 'application/pdf'
+            => 'picture_as_pdf',
+        (bool)preg_match('/^application\/(zip|x-zip|x-tar|x-rar|x-7z|gzip|x-bzip)/', $type)
+            => 'folder_zip',
+        (bool)preg_match('/^application\/(msword|vnd\.oasis\.opendocument\.text|vnd\.openxmlformats-officedocument\.wordprocessingml)/', $type)
+            => 'description',
+        (bool)preg_match('/^application\/(vnd\.ms-excel|vnd\.oasis\.opendocument\.spreadsheet|vnd\.openxmlformats-officedocument\.spreadsheetml)/', $type)
+            => 'table_chart',
+        (bool)preg_match('/^application\/(vnd\.ms-powerpoint|vnd\.oasis\.opendocument\.presentation|vnd\.openxmlformats-officedocument\.presentationml)/', $type)
+            => 'slideshow',
+        (bool)preg_match('/^text\/(html|xml|css|javascript)/', $type),
+        (bool)preg_match('/^application\/(json|xml|javascript)/', $type)
+            => 'code',
+        (bool)preg_match('/^text\//', $type)
+            => 'article',
+        typeIsAudio($type) => 'audio_file',
+        typeIsVideo($type) => 'video_file',
+        typeIsPicture($type) => 'image',
+        default => 'insert_drive_file',
+    };
+}
+
+/**
+ * @desc Return a short uppercase label from a mimetype (such as 'application/pdf' -> 'PDF')
+ */
+function mimeToLabel(string $type): string
+{
+    $subtype = explode('/', $type)[1] ?? '';
+
+    return strtoupper(match(true) {
+	$type === 'application/pdf'                                                      => 'pdf',
+        str_starts_with($subtype, 'vnd.openxmlformats-officedocument.wordprocessingml') => 'docx',
+        str_starts_with($subtype, 'vnd.openxmlformats-officedocument.spreadsheetml')    => 'xlsx',
+        str_starts_with($subtype, 'vnd.openxmlformats-officedocument.presentationml')   => 'pptx',
+        str_starts_with($subtype, 'vnd.oasis.opendocument.text')                        => 'odt',
+        str_starts_with($subtype, 'vnd.oasis.opendocument.spreadsheet')                 => 'ods',
+        str_starts_with($subtype, 'vnd.oasis.opendocument.presentation')                => 'odp',
+        str_starts_with($subtype, 'vnd.ms-excel')                                       => 'xls',
+        str_starts_with($subtype, 'vnd.ms-powerpoint')                                  => 'ppt',
+        $subtype === 'x-tar'                                                             => 'tar',
+        $subtype === 'x-7z-compressed'                                                  => '7z',
+        $subtype === 'x-bzip2'                                                           => 'bz2',
+        $subtype === 'x-rar'                                                             => 'rar',
+        $subtype === 'x-zip'                                                             => 'zip',
+        $subtype === 'msword'                                                            => 'doc',
+        default                                                                          => $subtype,
+    });
+}
--- a/app/Widgets/Chat/Chat.php
+++ b/app/Widgets/Chat/Chat.php
@@ -1516,8 +1516,10 @@ class Chat extends \Movim\Widget\Base
                         $message->body = '';
                     }
                 }
-            } elseif (isset($message->file) && $message->file->type != 'xmpp') {
-                $message->body = '';
+            } elseif (isset($message->file) && $message->file->type != 'xmpp' && !$message->file->preview) {
+                $view = $this->tpl();
+                $view->assign('file', $message->file);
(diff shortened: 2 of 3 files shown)


app/Widgets/Chat/chat.css
```diff
@@ -843,9 +843,9 @@ ul.list li>div.bubble.file div.file:not(:has(img)):not(:has(video)):not(:has(.au
     overflow: hidden;
     text-overflow: ellipsis;
     white-space: nowrap;
+    font-weight: 500;
```
divyaranjan1905: Now that I'm looking at it, maybe this was a bit too bold, let me know what you think

app/Helpers/UtilsHelper.php
```diff
@@ -1083,3 +1083,33 @@ function base64ToFingerPrint(string $base64): string
     $hex = unpack('H*', $buffer);
     return implode(' ', str_split(substr($hex[1], 2), 8));
 }
+
+/**
+ * @desc Return a Material Symbols icon name for a given mimetype
+ */
+function mimeToIcon(string $type): string
+{
+    if ($type === 'application/pdf') {
```
edhelas: You can use a match() there to not have plenty of if/else :) 

https://www.php.net/manual/en/control-structures.match.php

app/Widgets/Chat/chat.js
```diff
@@ -1160,11 +1160,18 @@ var Chat = {
             bubble.querySelector('div.bubble').classList.add('file');
             p.appendChild(Chat.getStickerHtml(data));
 
-        } else if (data.file != null && data.card === undefined && data.file.type !== 'xmpp' && data.retracted == false) {
-            bubble.querySelector('div.bubble').classList.add('file');
-            msg.appendChild(Chat.getFileHtml(data.file, data));
-        }
-
+	} else if (data.file != null && data.card === undefined && data.file.type !== 'xmpp' && data.retracted == false) {
```
edhelas: There is already a getCardHtml() function for that, you can directly inject the element as a card I think

divyaranjan1905: But the file bubble uses a `div.file` right, rather than a `ul.card` (which `getCardHtml` provides). So I thought it'd be structurally different, since I'm using a flex layout with `div.file`. Or am I missing something?

app/Helpers/UtilsHelper.php
divyaranjan1905: Apologies. Very rusty on PHP, after writing in it since a long time :)

app/Widgets/Chat/Chat.php
```diff
@@ -1516,9 +1516,18 @@ public function prepareMessage(&$message, $jid = null): array
...
-            }
+	    } elseif (isset($message->file) && $message->file->type != 'xmpp') {
+		if (!$message->file->preview) {
+		    $view = $this->tpl();
+		    $view->assign('file', $message->file);
+		    $view->assign('icon', mimeToIcon($message->file->type ?? ''));
+		    $view->assign('name', preg_replace('/^chat_/i', '', $message->file->name ?? ''));
+		    $view->assign('size', $message->file->cleansize ?? '');
+		    $view->assign('ext', strtoupper(pathinfo($message->file->name ?? '', PATHINFO_EXTENSION)));
+		    $message->file->cardHtml = $view->draw('_chat_file');
```
edhelas: Just put $message->card = there

app/Widgets/Chat/chat.js
edhelas: With $message->card in Chat.php you can just remove the whole block

app/Widgets/Chat/chat.css
```diff
@@ -803,10 +803,50 @@ ul.list li>div.bubble.file div.file[data-type="image/gif"]:after {
     pointer-events: none;
 }
 
-ul.list li>div.bubble.file div.file a span.host:after,
-ul.list li>div.bubble.file div.file a span.size:after {
-    content: '-';
-    margin: 0 0.5rem;
+ul.list li>div.bubble.file div.file:not(:has(img)):not(:has(video)):not(:has(.audio_player)) {
```
edhelas: With $message->card = in Chat.php you can just remove the whole CSS you added

app/Widgets/Chat/_chat_file.tpl
```diff
@@ -0,0 +1,5 @@
+<span class="primary icon gray"><i class="material-symbols">{$icon}</i></span>
```
edhelas: With $message->card = in Chat.php you just have to add 
```
<li class="block"> content </li> around those lines and you're done
```

edhelas:
Please squash all your commits in one commit only, easier to manage in the end :)
We're almost there

app/Widgets/Chat/Chat.php
```diff
@@ -1516,9 +1516,18 @@ public function prepareMessage(&$message, $jid = null): array
...
                     }
                 }
-            } elseif (isset($message->file) && $message->file->type != 'xmpp') {
-                $message->body = '';
-            }
+	    } elseif (isset($message->file) && $message->file->type != 'xmpp') {
+		if (!$message->file->preview) {
+		    $view = $this->tpl();
+		    $view->assign('file', $message->file);
+		    $view->assign('icon', mimeToIcon($message->file->type ?? ''));
```
edhelas: You're already passing $message->file as a parameter, you can directly use the type in the view

```
{$file->type ?? ''}
```

app/Widgets/Chat/Chat.php
```diff
@@ -1516,9 +1516,18 @@ public function prepareMessage(&$message, $jid = null): array
...
+		    $view->assign('name', preg_replace('/^chat_/i', '', $message->file->name ?? ''));
+		    $view->assign('size', $message->file->cleansize ?? '');
```
edhelas: Same there

app/Widgets/Chat/Chat.php
```diff
@@ -1516,9 +1516,18 @@ public function prepareMessage(&$message, $jid = null): array
...
+		    $view->assign('ext', strtoupper(pathinfo($message->file->name ?? '', PATHINFO_EXTENSION)));
```
edhelas: Using the file extension is not a good idea, extensions are not trustable regarding the content of the file

divyaranjan1905:
> Please squash all your commits in one commit only, easier to manage in the end :) We're almost there

Done :)

I added another helper to get the file type label instead of relying on the extension.

divyaranjan1905:
Looks like this, now:

image

divyaranjan1905:
We should probably do ellipsis on the file name properly to truncate it. But I have no idea about how to operate on the file name in this context. Also do you know why the timestamp on the card is weirdly positioned?

app/Widgets/Chat/_chat_file.tpl
```diff
@@ -0,0 +1,11 @@
+<li class="block">
+    <span class="primary icon bubble gray">
+        <i class="material-symbols fill">{$file->type|mimeToIcon}</i>
+    </span>
+    <div>
+        <p><a href="{$file->url}" target="_blank" rel="noopener noreferrer">{$file->name}</a></p>
+        {if="$file->cleansize"}
+            <p><span class="second">{$file->cleansize} · <span class="tag">{$file->type|mimeToLabel}</span></span></p>
```
edhelas: Remove the class=second and class=tag there, the CSS already take care of that

app/Widgets/Chat/Chat.php
```diff
@@ -1516,9 +1516,14 @@ public function prepareMessage(&$message, $jid = null): array
                         $message->body = '';
                     }
                 }
-            } elseif (isset($message->file) && $message->file->type != 'xmpp') {
-                $message->body = '';
-            }
+	    } elseif (isset($message->file) && $message->file->type != 'xmpp') {
+		if (!$message->file->preview) {
```
edhelas: Indentation...

edhelas: And maybe merge the if together?

app/Widgets/Chat/Chat.php
```diff
@@ -1516,7 +1516,11 @@ public function prepareMessage(&$message, $jid = null): array
                         $message->body = '';
                     }
                 }
-            } elseif (isset($message->file) && $message->file->type != 'xmpp') {
+            } elseif (isset($message->file) && $message->file->type != 'xmpp' && !$message->file->preview) {
+                $view = $this->tpl();
+                $view->assign('file', $message->file);
+                $message->card = $view->draw('_chat_file');
+            } else {
```
edhelas: No... no else there, the body is cleared when it was handled as a card :p

divyaranjan1905: Oops, sorry! Fixed. I thought it might be good to leave the body as empty in some other case.

edhelas:
I'll take care of the rest in another commit.

PR #1547, opened 2026-03-15
Update de.po

bugtrapper:
Streamlined addressing in German (du), added German translations, added translator name


edhelas:
The translations are handled in Transifex https://app.transifex.com/movim/movim/

PR #1568, opened 2026-04-06
MuJi Refactor & Fixes

divyaranjan1905:
This PR fixes several things in MuJi, namely:

- call participants count in header not updating after someone leaves
- one participant leaving leads to broken call widget for other participants
- non-participants sometimes do not see that there is a call going on, thus accidentally start a new call

The primary fix is that now users can reliably leave the call without affecting others. Additionally now the call initiator has two buttons: one to leave the call silently (orange, near header), another to end the entire call (red, in the widget).


divyaranjan1905:
The header counter still somtimes doesn't update when somebody leaves, I might be missing to update it somewhere else.

app/Widgets/Visio/Visio.php
```diff
@@ -474,32 +484,49 @@ public function ajaxLeaveMuji(string $mujiId)
...
+                ->request();
+        }
 
+        if ($this->currentCall()) {
             $this->currentCall()->stop($muji->jidfrom, $muji->id);
-            $this->rpc('MovimJingles.terminateAll');
         }
+
+        $this->rpc('MovimJingles.terminateAll');
+        // Update the header after leaving
```
edhelas: You can remove the comment there

src/Moxl/Xec/Payload/Presence.php
```diff
@@ -25,23 +25,27 @@ public function handle(?\SimpleXMLElement $stanza = null, ?\SimpleXMLElement $pa
...
+            $muji = $this->me->session->mujiCalls()
+                ->where('muc', $jid['jid'])
+                ->first();
+
             if (
-                linker($this->me->session->id)->currentCall->isStarted()
-                && linker($this->me->session->id)->currentCall->mujiRoom == $jid['jid']
+                $muji
+                && $stanza->muji
+                && $stanza->muji->attributes()->xmlns == 'urn:xmpp:jingle:muji:0'
```
edhelas: Here you can simplify the writing to `$stanza->muji?->attribute()...` with a null safe operator https://wiki.php.net/rfc/nullsafe_operator

src/Moxl/Xec/Payload/Presence.php
```diff
@@ -25,23 +25,27 @@ public function handle(?\SimpleXMLElement $stanza = null, ?\SimpleXMLElement $pa
             $presence = DBPresence::findByStanza($this->me, $stanza);
             if (!$presence->set($this->me, $stanza)) return;
 
+            // Treat MuJi presence as room-authoritative so a fresh session
+            // can discover an ongoing call even if the browser is not already in it.
+            $muji = $this->me->session->mujiCalls()
```
edhelas: This is doing a SQL request for each and every incoming presence, that's why it was inside the if() bellow.

app/Widgets/Visio/Visio.php
```diff
@@ -474,32 +484,49 @@ public function ajaxLeaveMuji(string $mujiId)
...
-
-                // If we were the inviter, we also retract the call
-                $participant = $muji->participants->firstWhere('jid', $muji->jidfrom . '/' . $resource);
-                if ($participant && $participant->inviter) {
-                    $retract = $this->xmpp(new Retract);
-                    $retract->setTo($muji->jidfrom)
-                        ->setId($muji->id)
-                        ->request();
-                }
+        if (!$muji) {
```
edhelas: For a simple return you can put everything in one line and remove the brackets

app/Widgets/Visio/Visio.php
```diff
@@ -484,32 +484,27 @@ public function ajaxLeaveMuji(string $mujiId)
...
+        }
+
+        if ($this->currentCall()) {
             $this->currentCall()->stop($muji->jidfrom, $muji->id);
-            $this->rpc('MovimJingles.terminateAll');
         }
+
+        $this->rpc('MovimJingles.terminateAll');
+        // Update the header after leaving
+        $this->rpc('Chat_ajaxGetHeader', $muji->jidfrom, true);
```
divyaranjan1905, line 503: @edhelas How come this isn't properly updating the participant counter in the header? I checked it in `_chat_header.tpl` too, but it's still not updating reliably.

edhelas:
Please rebase and merge your commits :)

app/Widgets/Chat/_chat_header.tpl
```diff
@@ -78,11 +78,23 @@
...
                     {if="$muji = $conference->currentMuji()"}
-                        <button class="button oppose color red"
-                                onclick="Visio_ajaxLeaveMuji('{$muji->id}')">
-                            <i class="material-symbols">{$muji->icon}</i>
-                        </button>
+                        {if="$muji->joined"}
+                            <button class="button oppose color orange"
+                                    onclick="Visio_ajaxLeaveMuji('{$muji->id}')"
+                                    title="{$c->__('visio.leave_call')}">
+                                <i class="material-symbols">{$muji->endIcon}</i>
```
edhelas: A think a different icon would be cleared, something like phone_forwarded

app/Widgets/Chat/_chat_header.tpl
```diff
@@ -78,11 +78,23 @@
...
+                            </button>
+                        {else}
+
```
edhelas: Useless carriage return

app/Widgets/Visio/Visio.php
edhelas: participants != presences, when you leave a call you are still part of the participants list because you accepted the call

app/MujiCall.php
```diff
@@ -66,4 +66,16 @@ public function getendIconAttribute()
     {
         return $this->video ? 'videocam_off' : 'call_end';
     }
+
+    public function getIsInitiatorAttribute(): bool
+    {
+        $linker = linker($this->session_id);
+        if (!$linker || !$linker->user) return false;
+
+        $myJid = $linker->user->id;
```
edhelas: No need to declare it here, just use it bellow

public/scripts/movim_visio.js
```diff
@@ -44,7 +44,12 @@ var MovimVisio = {
         visio.dataset.jid = jid;
         visio.dataset.type = (withVideo) ? 'video' : 'audio';
         visio.dataset.muji = isMuji ? 'true' : 'false';
+        visio.dataset.mujiOwner = canEndForEveryone ? 'true' : 'false';
```
edhelas: Not a fan of using capital letters in datase, "mujiinviter"

public/scripts/movim_visio.js
```diff
@@ -30,7 +30,7 @@ var MovimVisio = {
         MovimVisio.localAudio = document.getElementById('local_audio');
     },
 
-    init: function (fullJid, jid, id, withVideo, isMuji) {
+    init: function (fullJid, jid, id, withVideo, isMuji, canEndForEveryone) {
```
edhelas: Just call it isInviter, its more generic

src/Moxl/Xec/Payload/Presence.php
```diff
@@ -26,22 +26,23 @@ public function handle(?\SimpleXMLElement $stanza = null, ?\SimpleXMLElement $pa
...
+                // can discover an ongoing call even if the browser is not already in it.
                 $muji = $this->me->session->mujiCalls()
                     ->where('muc', $jid['jid'])
                     ->first();
 
-                if ($muji) {
-                    $this->pack($muji);
-                    $this->method('muji_event');
-                    $this->deliver();
+                $this->pack($muji);
```
edhelas: $muji could be null there, please keep the test

app/Widgets/Visio/Visio.php
divyaranjan1905: Ah. The counter in the header should be presences though right? I think it might be confusing to have both in the UI. I think we should only expose the presences in the UI. But we already do here, right? Then why doesn't it seem to work...or maybe I'm misinterpreting it.

https://github.com/movim/movim/blob/2ef7c49c6b03ba8b2487ddab1e2648ec08015b63/app/Widgets/Chat/_chat_header.tpl#L150

Issue #1576, opened 2026-04-15
Movim closes the XMPP stream before without sending anything during auth


truenicoco:
Using the Containerfile to launch movim and connecting to a local prosody server, it often [1] fails. The web UI shows the "connected" toast then refreshes.

In `errors.log` I can see an interesting `InvalidArgumentException:  is not a valid ini size in /var/www/html/vendor/react/http/src/Io/IniUtil.php (line 26)`

<details>

```
[2026-04-15T16:36:00.968459+00:00] movim.ERROR: InvalidArgumentException:  is not a valid ini size in /var/www/html/vendor/react/http/src/Io/IniUtil.php (line 26)
Trace
#0 /var/www/html/vendor/react/http/src/HttpServer.php(347): React\Http\Io\IniUtil::iniSizeToBytes()
#1 /var/www/html/vendor/react/http/src/HttpServer.php(231): React\Http\HttpServer->getMaxRequestSize()
#2 /var/www/html/workers/templater.php(37): React\Http\HttpServer->__construct()
#3 {main}  
[2026-04-15T16:36:40.531190+00:00] movim.ERROR: InvalidArgumentException:  is not a valid ini size in /var/www/html/vendor/react/http/src/Io/IniUtil.php (line 26)
Trace
#0 /var/www/html/vendor/react/http/src/HttpServer.php(347): React\Http\Io\IniUtil::iniSizeToBytes()
#1 /var/www/html/vendor/react/http/src/HttpServer.php(231): React\Http\HttpServer->getMaxRequestSize()
#2 /var/www/html/workers/templater.php(37): React\Http\HttpServer->__construct()
#3 {main}  
[2026-04-15T16:37:06.754662+00:00] movim.ERROR: Allowed memory size of 536870912 bytes exhausted (tried to allocate 400560128 bytes) in /var/www/html/src/Movim/Daemon/Linker.php (line 245)
  
[2026-04-15T16:39:38.068451+00:00] movim.ERROR: Allowed memory size of 536870912 bytes exhausted (tried to allocate 400560128 bytes) in /var/www/html/src/Movim/Daemon/Linker.php (line 245)
```

</details>

I patched `Movim/Daemon/Linker.php` to discriminate between "error" and "close" events like this:

```php
        $this->connection->on('error', fn() => $this->onError($this->connection));
        $this->connection->on('close', fn() => $this->onClose($this->connection));
...
    private function onError(Connection $conn)
    {
        logOut("there is an error!");
[...]

PR #1579, opened 2026-04-22
Improve container

truenicoco:
Some small improvements while trying to investigate https://github.com/movim/movim/issues/1576

I marked this as WIP because I want to push some more improvements in here. @edhelas would you agree in principle with the following changes, if I ever submit them?

- [ ] ~Disable SSL entirely. The practice is usually to have containers do PLAIN and put them behind a reverse proxy. Moreover, here, the container is for development more than deployment, so we care even less about SSL.~
- [ ] ~Have postgresql run in the same container too. We have a frankenstein-y situation here, where several services are spawned in the movim container, but then we need postgresql in another one too. For deployment it's usually considered good practice to have one process per container, but I think we're aiming at a development-oriented container here, so I think it would even be more convenient to let go of ssl all along?~
- [x] Have it built and pushed to ghcr.io in github actions

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -1,0 +1,18 @@
+name: Build and publish a Docker image to ghcr.io
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  docker_publish:
+    runs-on: "ubuntu-24.04"
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build and publish a Docker image for ${{ github.repository }}
+        uses: macbre/push-to-ghcr@master
+        with:
+          dockerfile: Containerfile
+          image_name: ${{ github.repository }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
--- a/Containerfile
+++ b/Containerfile
@@ -2,17 +2,19 @@ FROM docker.io/serversideup/php:8.4-fpm-nginx-alpine3.21
 
 USER root
 
+# PHP
+RUN install-php-extensions imagick gd bcmath
+
 # S6
 COPY ./etc/s6-overlay/s6-rc.d/movim-migrations/ /etc/s6-overlay/s6-rc.d/movim-migrations/
 COPY ./etc/s6-overlay/s6-rc.d/movim-daemon/ /etc/s6-overlay/s6-rc.d/movim-daemon/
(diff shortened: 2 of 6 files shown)


edhelas:
- I don't care about SSL or not but in any case Movim _need_ to have SSL enabled when working so if it is not done in the container we need a documentation to put it on top + a reverse proxy to do it. What is the plan there, to have a "one command that does everything" or have a minimal container that does "just it needs" and an ecosystem around that to do the rest ?
- No, containers means we explode things the best we could, its not because we must have php-fpm and nginx/Apache in the same container that we must do everything like that. What is the plan there, to have a "one command that does everythin" or have a minimal container that does "just it needs" ? :p
- I don't know what it is, I guess if it help do it

truenicoco:
> * Movim _need_ to have SSL

image

- [x] doubt :P

Where is SSL needed? In any case, SSL support right now is "generate a new self-signed certificate at each startup" which is very meh. At the very least, we should have a re-usable self-signed certificate, so it can easily be added to the system trust thing to avoid having to "accept security exceptions" in the browser every time we restart the container. I can try to implement that, but again, are you sure SSL is mandatory? It seems to be working without it for me right now. Looks like I can disable it with an env var in slidge's `podman-compose.yml` though, so I can also do that.

> containers means we explode things the best we could

Hmmm OK, if you want to go there, then the way forward is probably to try and split the movim daemon in its own container too. I don't think I want to go there as I don't need that (as opposed to "everything in one container" which would make _my_ life easier ^^), but it's good that I know.

> I don't know what it is, I guess if it help do it

GHCR = github container registry. In slidge's `podman-compose.yml` I'll be able to use `image: "ghcr.io/movim/movim:latest"` and BOOM, no need to maintain anything else for me. :)

edhelas:
- WebRTC features and some others doesn't work without HTTPS, https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia for example. Let's not do webpages like its 2003 again :p

truenicoco:
OK, I think we should actually point to the docs of the base image you chose: https://serversideup.net/open-source/docker-php/docs/deployment-and-production/configuring-ssl - it should be flexible enough for most use case. For slidge's `podman-compose.yaml` I just need to set `SSL_MODE: off`.

PR #1585, opened 2026-05-08
Remove the invitation based Muji system and drop the XEP-0482: Call I…

edhelas:
…nvites


divyaranjan1905:
We no longer get the message of "A Conference call has started" and the ending message. Is this expected? It's these messages:
image

edhelas:
Yes, because it's a totally different flow now, those messages might come back later

divyaranjan1905:
> Yes, because it's a totally different flow now, those messages might come back later

We can still implement though, right?

divyaranjan1905:
Sometimes when A is not able to connect with B, but can hear/listen C, there isn't any way to figure out that, and people talk past each other. How can we make that a bit more stable? It corrects after rejoining, but it happens quite often than we'd like to.

divyaranjan1905:
> > Yes, because it's a totally different flow now, those messages might come back later
> 
> We can still implement though, right?

Nevermind. I understood the change. We don't need the messages anymore.

app/Widgets/Chat/_chat_header.tpl
```diff
@@ -76,28 +76,17 @@
...
-                            <button class="button oppose color blue {if="$incall"}disabled{/if}"
-                                    onclick="Visio_ajaxJoinMuji('{$muji->id}', {if="$muji->video"}true{else}false{/if});">
-                                <i class="material-symbols {if="$muji->joined"}blink{/if}">
-                                    {$muji->icon}
-                                </i>
-                            </button>
-                        {/if}
+                        <button class="button oppose color blue {if="$incall"}disabled{/if}"
+                                onclick="Visio_ajaxJoinMuji('{$conference->conference}', {if="$conference->mujiPresences->first()->hasVideoMuji()"}true{else}false{/if});">
+                            <i class="material-symbols">call</i>
```
divyaranjan1905, line 88: I think we should replace `call` here with `voice_chat` and similarly for `videocam` with `video_chat`. Since now we're following Telegram/Mumble/Discord conventions, and we're no longer ringing every single peer.

PR #1593, opened 2026-05-17
fix HTML escaping in encrypted messages

balejk:
The original fix only sanitized the plaintext from XSS-unsafe HTML
elements, however it is not desirable to render any HTML injected by the
sender.

Also, the original fix performed escaping after scanning the message for
file uploads which would also escape the safe HTML displaying the file
download icon (although only at the Safari variant).

---

I have not tested this as I don't have a Movim instance set up.


edhelas:
Why did you used the Option object for this ?

balejk:
I found it as a tip online, probably any other object would work too,
the point is that the string passed to the constructor is escaped by the
browser. If there is any more straightforward way to perform the
escaping, I will be happy to change it.

Issue #1551, opened 2026-03-25
Allow enabling video & screensharing from voice calls


divyaranjan1905:
In Dino you can initiate a voice call and then enable your video. I think this is useful when you aren't sure if you want to share your video but at somepoint in the call you need to. Having to restart a new call is cumbersome.

And sometimes you only need to do screensharing on an audio call without having to first share your video.

divyaranjan1905:
This is not planned for now until we are done with 0.33.

PR #1597, opened 2026-05-20
Allow audio/video/screen streams to be dynamicaly added in a session

edhelas:
Detect properly if a new media is a screen sharing one in updateContent
Get rid of Visio type as things are dynamics now


edhelas:
Fix #1551

PR #1623, opened 2026-08-25
Fix iOS Safari image upload sending empty request body

### LINKED ISSUE #1621: Image upload silently fails on iOS Safari (empty request body)
Environment:

Client: Movim web (mov.im instance)
Browser: Safari on iOS 26.6 (iPhone)
XMPP server: friend-hosted (not movim.eu backend)
Same account works fine for images in Monal (native iOS XMPP client)

Steps to reproduce:

Open a chat on mov.im in iOS Safari
Tap the upload/attach button
Select an image from the photo picker
Nothing happens — no preview, no error, no message sent

Observed behavior (via Web Inspector, Mac + iPhone tethered):

No JS errors/exceptions appear in console
A POST request does fire to https://mov.im/upload/<uuid>
Server responds 200 OK
However, the request has Content-Length: 0 — the multipart/form-data body is empty, so no file bytes are actually sent
Full request headers (session cookie redacted):
  POST /upload/890ff2d6-f5bf-4631-824b-8f62e54b9868 HTTP/1.1
  Content-Type: multipart/form-data; boundary=----WebKitFormBoundaryYDqW2KS3axZQ3GD3
  Content-Length: 0
  Origin: https://mov.im
  User-Agent: Mozilla/5.0 (iPhone; CPU iPhone OS 18_7 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/26.6 Mobile/15E148 Safari/604.1

(Note: "Version/26.6" is the actual Safari/iOS version — the "CPU iPhone OS 18_7" segment of the UA string is a known Apple compatibility quirk and doesn't reflect the real OS version.)

Expected behavior:
Selected image should be included in the upload request body and sent successfully, as it does on desktop browsers and in Monal (native app, same account/server).

Suspected cause:
Likely a known iOS Safari quirk where the File/Blob reference from <input type="file"> becomes invalid if not appended to FormData synchronously within the same event-loop tick as the user's change event (e.g. if the code awaits something — like a FileReader preview step — before building the FormData). Worth checking the upload handling JS (historically app/widgets/Upload/upload.js) for any async gap between file selection and FormData.append.

monopatis:
The original File reference from <input type='file'> becomes invalid on iOS Safari after any async gap (FileReader, Image load). When the image is small (<= SMALL_PICTURE_LIMIT), the file was passed through two async hops before being appended to FormData, resulting in Content-Length: 0 uploads.

Now the small-image path also creates a fresh Blob from the canvas via canvas.toBlob(), matching the large-image path behaviour.

Fixes #1621

tested in iOS 26.6.1
image

### DIFF
--- a/app/Widgets/Upload/upload.js
+++ b/app/Widgets/Upload/upload.js
@@ -194,7 +194,20 @@ var Upload = {
                 ctx = Upload.canvas.getContext("2d");
                 ctx.drawImage(image, 0, 0, image.naturalWidth, image.naturalHeight);
 
-                Upload.prepare(file);
+                // Always create a fresh Blob from the canvas for iOS Safari.
+                // The original File reference becomes invalid after async gaps,
+                // producing Content-Length: 0 uploads on Safari.
+                if (typeof Upload.canvas.toBlob == 'function') {
+                    Upload.canvas.toBlob(
+                        function (blob) {
+                            Upload.prepare(blob);
+                        },
+                        file.type,
+                        1.0
+                    );
+                } else {
+                    Upload.prepare(file);
+                }
             }
         });
         image.src = src;


edhelas:
Ah nice :) ! I'll read it tomorrow and merge it if it looks fine.

Thanks a lot for the contribution !

edhelas:
Thx a lot for the contribution ! :heart_hands:

PR #1628, opened 2026-09-10
Preserve the current page number when pushing news soft state

KevinPolez:
### Problem

When browsing `/news` on a page > 0, clicking a post ("Read more") and
then using the browser back button returns the user to `/news` (page 0)
instead of the page they were on.

### Cause

The three `ajaxHttpGet*` handlers in `app/Widgets/Menu/Menu.php` push a
soft history state with `$this->route('news')` — without the current
page number. Even though the client-side `MovimUtils.reload()` initially
pushes `/news/N`, this server-side RPC then overwrites the URL with
`/news`, and that entry is what the browser goes back to.

### Fix

Pass the current `$page` to `route('news', ...)` so the pushed URL
matches the page actually being displayed. Uses `$page ?: null` to keep
the default URL as `/news` for page 0 (aligning with the `$params` default
of `Route::urlize()`, rather than depending on the loose `!= null`
comparison inside `urlize()`).

### Reproduction

1. Open `/news` and click "Next" a few times (URL becomes `/news/2`).
2. Click "Read more" on any post.
3. Press the browser back button.

Before: URL and content reset to page 0.
After: back to `/news/2` where the user was.

Same fix applied to the three tabs: All, Communities, Contacts.


edhelas:
A small but very valuable fix :) ! Thanks.
I'll review it and try to merge it asap ;)

edhelas:
Everything is good! Thanks for the fix :)

PR #1629, opened 2026-09-13
Stop rendering the featured image twice on WordPress feeds

KevinPolez:
Feeds from WordPress-based news sites often publish the featured image twice: once as an <enclosure>, once at the top of the <content:encoded> HTML. Movim rendered both sources, so the card showed the same image on top of itself.

The leading content image is already extracted as $extra in Post::set and passed to setAttachments. When it matches an enclosure picture, strip it from contentcleaned once — the stored HTML stays clean and getContent is untouched. Wrappers like <figure><a></a> are absorbed so no empty shell is left behind.

WordPress serves the same image under different size suffixes such as -1280x853, so URLs are compared after stripping that pattern. Distinct filenames in the same directory are left alone (tested on Korben, whose feed exposes two different images).

Tested on feeds that showed the duplicate before the patch: JournalDuGeek, Presse-Citron.


edhelas:
This is too hacky and Wordpress specific to me, sorry

edhelas:
I'd prefer that Movim doesn't "fix" all the hacky Atom items out there, its not its role. Movim parse Atom 1.0 and display it as it is.

If Wordpress is generating bad RSS items, they are converted to Atom 1.0 using https://github.com/edhelas/feedcleaner , I'd prefer those fixes to be in this library :)

PR #1630, opened 2026-09-14
Remove jabber:client namespace from Galener stanza creation to support Prosody

delonium:
According to XEP-0114, external components should send stanzas with the `jabber:component:accept` namespace. Right now, Galener constructs stanzas with the `jabber:client` namespace, which is incorrect. Removing the namespace entirely allows the stanza to inherit the namespace from the authenticated external component stream.

I discovered this due to testing with Prosody, as the STUN/TURN discovery feature of Galener would be rejected by Prosody. This single line change allows the blue "Enable Wide Conference Call" banner to show up on my Movim instance while logged in with an account on my Prosody server. Previously, the feature was silently not present.


delonium:
Okay, after a quick test this morning, the sfu is not getting added to the space when I enable the SFU feature. Prosody is complaining about the namespace when it's present in a `presence` stanza as well. It looks like I need to be more thorough.

delonium:
Changing the `presence` stanza creation now allows the sfu user to join my space, but Movim still thinks the feature is not enabled and continues showing the blue banner. Will continue investigating.

delonium:
Update: This line in [app/Conference.php](https://github.com/movim/movim/blob/master/app/Conference.php#L56) is what is signaling that the SFU feature is not being detected. I've tracked the problem down to the `handle` function in [app/Workers/Galener/Events/Presence.php](https://github.com/movim/movim/blob/master/app/Workers/Galener/Events/Presence.php#L14), this never calls `$conference->xmppAddMember`, which would then actually generate the presence with coin in the `generatePresence` by setting `Conference`'s `connected` member to `true`: [app/Workers/Galener/Conference.php](https://github.com/movim/movim/blob/master/app/Workers/Galener/Conference.php#L236). My next step is trying to figure out why `$conference->xmppAddMember` is never fired, so I ventured in logging some useful info here. I'll be honest and say I did ask an LLM as to what would be useful to log here (I'm still getting used to this codebase and XMPP's technical side in general), but the results are interesting:

```
[2026-09-17T07:00:58.492571+00:00] movim.DEBUG: GALENER presence event: from=0340d468-ba44-48f4-857a-2a951b71bd03@groups.XMPP_SERVER/sfu to=sfu.XMPP_SERVER conf=yes [] []
[2026-09-17T07:00:58.495698+00:00] movim.DEBUG: GALENER item: jid=sfu.XMPP_SERVER affiliation=admin role=moderator status110=1 to=sfu.XMPP_SERVER sfuJid=5de3bfc0536dbe820b821aec4dcb1bcbdeba9f83@sfu.XMPP_SERVER [] []
[2026-09-17T07:00:58.501836+00:00] movim.DEBUG: GALENER presence event: from=0340d468-ba44-48f4-857a-2a951b71bd03@groups.XMPP_SERVER/sfu.XMPP_SERVER_J84Jrp to=5de3bfc0536dbe820b821aec4dcb1bcbdeba9f83@sfu.XMPP_SERVER conf=yes [] []
[2026-09-17T07:00:58.504652+00:00] movim.DEBUG: GALENER presence with NO muc#user item: <?xml version="1.0"?> <presence to="5de3bfc0536dbe820b821aec4dcb1bcbdeba9f83@sfu.XMPP_SERVER" type="error" from="0340d468-ba44-48f4-857a-2a951b71bd03@groups.XMPP_SERVER/sfu.XMPP_SERVER_J84Jrp"><error by="0340d468-ba44-48f4-857a-2a951b71bd03@groups.XMPP_SERVER" type="auth"><registration-required xmlns="urn:ietf:params:xml:ns:xmpp-stanzas"/></error></presence>  [] []
```

I'll dissect this output: First, the presence event is always printed (XMPP_SERVER is my redacted server root domain). Next, either this line in [app/Workers/Galener/Events/Presence.php](https://github.com/movim/movim/blob/master/app/Workers/Galener/Events/Presence.php#L17) will handle the SFU user (the `GALENER item` part) or not (the `GALENER presence with NO muc#user item` part).

Basically, what I'm reading from this is that the bare JID of the SFU is correctly being set as an admin, but the other user (who I suppose is actually used to advertise the SFU feature and would bear the coin xml) is not authorized to be registered. This could be another tangible Prosody quirk! I feel like I'm getting closer to figuring this out.

delonium:
Made another step! It seems that Prosody is more strict than ejabberd in another regard: affiliation change to admin. Technically according to [XEP-0045](https://xmpp.org/extensions/xep-0045.html#affil-change) only *owners* can change an unaffiliated user into an admin. The bare SFU JID is set as an admin, so it can't promote the advertising user to an admin. I simply changed the affiliation stanza to promote the SFU advertising user to member instead, and now the SFU feature is detected!

I can join an SFU-powered call on Prosody... however, I get this error when I join/leave the call:

```
[2026-09-19T00:04:33.838875+00:00] movim.ERROR: simplexml_load_string(): namespace error : xmlns: '{https://movim.eu}conference_started_at' is not a valid URI in /var/www/movim/app/Presence.php (line 153)
```

Also, my user doesn't leave the participant list when I click the leave button (my nickname and avatar still show up under the space room). Perhaps this error is causing state to act weird.

edhelas:
So, is it good now :) ?

delonium:
I'm getting close :+1: I crammed some logs in Galener's `XMPPHandler.php` and realized that my client was sending stanzas like this that went unhandled:

```
<?xml version="1.0"?> <message from="delonium@XMPP_SERVER/Movim.XTczFm~Rx0NmBSkMjK7" type="chat" lang="en" to="580074769302e2967094969406c68d192a176ad0@sfu.XMPP_SERVER"><retract id="a715d121-bf1f-4d46-b0cd-4c20c610bfe4" xmlns="urn:xmpp:jingle-message:0"><reason xmlns="urn:xmpp:jingle:1"><cancel/><text>Retracted</text></reason></retract></message>
```

So I added a new event that's copy-paste of `JingleFinish.php` to remove the connection upon retraction, then now my user can join/leave cleanly by himself. Will force-push this here soon, then I will test with multiple users to make sure the connections get added/removed from the Galene conference. If this works, I think it's ready!

